PR #56, opened 2019-03-04
Add more linters

SimonBarendse:
Add the following linters:

  - unused
  - gosimple
  - stylecheck
  - structcheck
  - varcheck
  - interfacer
  - unconvert
  - ineffassign
  - deadcode
  - gocyclo
  - typecheck
  - depguard
  - unparam
  - nakedret
  - gofmt
  - misspell
  - prealloc
  - goconst

The last four found issues, which I've fixed.

### DIFF
--- a/internals/crypto/ciphertext.go
+++ b/internals/crypto/ciphertext.go
@@ -104,9 +104,11 @@ func newEncodedCiphertextMetadata(metadataList map[string]string) encodedCiphert
 	metadata := ""
 
 	// Sort all the keys of the metadataList so that metadata is always in alphabetical order.
-	var keys []string
+	keys := make([]string, len(metadataList))
+	i := 0
 	for k := range metadataList {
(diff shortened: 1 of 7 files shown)


internals/crypto/ciphertext.go
```diff
@@ -104,9 +104,11 @@ func newEncodedCiphertextMetadata(metadataList map[string]string) encodedCiphert
 	metadata := ""
 
 	// Sort all the keys of the metadataList so that metadata is always in alphabetical order.
-	var keys []string
+	keys := make([]string, len(metadataList))
+	i := 0
 	for k := range metadataList {
```
mackenbach: The parameter `metadataList` is confusingly named. Because it's a `list` I exptected it to be a slice and expected to see simpler iteration logic. Upon inspection I realized it's a map, hence the conversion. Not blocking at all, but would be nice to name it intuitively.

PR #57, opened 2019-03-06
Rename metadataList to metadataMap

SimonBarendse:
Because it is a map, not a list.


internals/crypto/ciphertext.go
```diff
@@ -100,13 +100,13 @@ func (ec encodedCiphertext) metadata() (encodedCiphertextMetadata, error) {
 
 // newEncodedCiphertextMetadata creates a new encodedCiphertextMetadata from a map of metadata.
 // Input of {"param": "foo", "second": "bar"} outputs "param=foo,second=bar".
-func newEncodedCiphertextMetadata(metadataList map[string]string) encodedCiphertextMetadata {
+func newEncodedCiphertextMetadata(metadataMap map[string]string) encodedCiphertextMetadata {
```
mackenbach: Actually, this variable would better be called `metadata` and then rename the variable below to `result` or something

SimonBarendse: You are right. I've renamed it.

PR #58, opened 2019-03-07
Create encode and decode functions for ciphertexts

SimonBarendse:
Created encode and decode functions to separate the encoding from marshaling to JSON.

### DIFF
--- a/internals/crypto/rsa.go
+++ b/internals/crypto/rsa.go
@@ -358,8 +358,8 @@ type CiphertextRSAAES struct {
 	rsa CiphertextRSA
 }
 
-// MarshalJSON encodes the ciphertext in a string.
-func (ct CiphertextRSAAES) MarshalJSON() ([]byte, error) {
+// Encode encodes the ciphertext in a string.
+func (ct CiphertextRSAAES) Encode() string {
 	data := base64.StdEncoding.EncodeToString(ct.aes.Data)
 
 	metadata := newEncodedCiphertextMetadata(map[string]string{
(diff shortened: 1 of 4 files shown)


internals/crypto/rsa.go
```diff
@@ -358,73 +358,88 @@ type CiphertextRSAAES struct {
...
+func (ct CiphertextRSAAES) EncodeToString() string {
 	data := base64.StdEncoding.EncodeToString(ct.aes.Data)
 
 	metadata := newEncodedCiphertextMetadata(map[string]string{
 		"nonce": base64.StdEncoding.EncodeToString(ct.aes.Nonce),
 		"key":   base64.StdEncoding.EncodeToString(ct.rsa.Data),
 	})
 
-	return json.Marshal(fmt.Sprintf("%s$%s$%s", algorithmRSAAES, data, metadata))
+	return fmt.Sprintf("%s$%s$%s", algorithmRSAAES, data, metadata)
```
mackenbach: Tiny note: using `Sprintf` for simple string concatenation is pretty 'slow' when compared to `algorithmRSAAES+"$"+data+"$"+metadata` or using something like `StringBuilder`.

Totally not a priority right now, but it's worth checking out:

https://gist.github.com/dtjm/c6ebc86abe7515c988ec

https://stackoverflow.com/questions/1760757/how-to-efficiently-concatenate-strings-in-go/47798475#47798475

internals/crypto/rsa.go
```diff
@@ -358,73 +358,88 @@ type CiphertextRSAAES struct {
...
-		return nil
-	}
+// MarshalJSON encodes the ciphertext in JSON.
+func (ct CiphertextRSAAES) MarshalJSON() ([]byte, error) {
+	return json.Marshal(ct.EncodeToString())
+}
 
-	encoded, err := newEncodedCiphertext(s)
+// DecodeCiphertextRSAAESFromString decodes an encoded ciphertext string to an CiphertextRSAAES.
+func DecodeCiphertextRSAAESFromString(ct string) (CiphertextRSAAES, error) {
```
mackenbach: Consider renaming `ct` to `s`

internals/crypto/rsa.go
SimonBarendse: :+1: 

Let's pick this up in a new PR, as this PR does not change the status-quo.

PR #60, opened 2019-03-08
Prettify README.md

mackenbach:
This improves the first impression of the README.md a bit, imo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+<p align="center">
+    <img src="https://secrethub.io/img/secrethub-gopher.png" alt="Gopher" width="160px"/>
+</p>
+
 # Go client for SecretHub
 
 [![GoDoc](http://img.shields.io/badge/godoc-reference-blue.svg)][godoc]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,15 +1,19 @@
...
 
 [![GoDoc](http://img.shields.io/badge/godoc-reference-blue.svg)][godoc]
 [![Travis CI](https://travis-ci.org/secrethub/secrethub-go.svg?branch=master)][travis-ci]
 [![GolangCI](https://golangci.com/badges/github.com/secrethub/secrethub-go.svg)][golang-ci]
 [![Go Report Card](https://goreportcard.com/badge/github.com/secrethub/secrethub-go)][goreportcard]
 
-The official [SecretHub][secrethub] Go client library.
+[SecretHub][secrethub] is a developer tool to help you keep database passwords, API tokens, and other secrets out of IT automation scripts. 
+
+Packages in `secrethub-go` provide clients for various SecretHub APIs.
```
SimonBarendse: Do we want to use plural (clients) here? If we are consistent with the names used in the code, we provide a single client that provides multiple services.

Also, there's 1 package that provides these services: `github.com/secrethub/secrethub-go/pkg/secrethub`. So maybe we should use singular for packages as well.

We could make it:
```
`secrethub-go` provides a client for various SecretHub APIs.
```

PR #62, opened 2019-03-18
Create a client.Me() service to retrieve own user and repos

SimonBarendse:
I've created a service that aliases `client.Repos().ListMine()` and `client.Users().Me()`.

I'll create another PR that adds `client.Me().SendVerficationEmail()`.

### DIFF
--- a/pkg/secrethub/me.go
+++ b/pkg/secrethub/me.go
@@ -1,0 +1,33 @@
+package secrethub
+
+import "github.com/secrethub/secrethub-go/internals/api"
+
+// MeService handles operations on the authenticated account.
+type MeService interface {
+	// Repos retrieves all repositories of the current user.
+	Repos() ([]*api.Repo, error)
+	// User retrieves the current users details.
+	User() (*api.User, error)
+}
+
+type meService struct {
+	repoService RepoService
+	userService UserService
+}
+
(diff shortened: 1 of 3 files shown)


pkg/secrethub/me.go
```diff
@@ -0,0 +1,33 @@
...
+
+func newMeService(repoService RepoService, userService UserService) MeService {
+	return meService{
+		repoService: repoService,
+		userService: userService,
+	}
+}
+
+// Repos retrieves all repositories of the current user.
+func (ms meService) Repos() ([]*api.Repo, error) {
```
mackenbach: Should this be `Repos` or `ListRepos`? Using `Repos` kind of feels like you would expect a `RepoService` that's configured to work for your own repos or something?

Not sure though, also like the simplicity of a simple `Repos` call. Also, what else would we add? Not `Repo` (singular), maybe `Orgs`, but not `Org` etc. 

Anyway, your call both are good.

SimonBarendse: I prefer `ListRepos`, because we use verbs for all other actions as well. Will change it.

SimonBarendse: By the same logic, `User` will become `GetUser`.

PR #63, opened 2019-03-18
 Add SendVerificationEmail function to the MeService 

SimonBarendse:
~~#62 should be merged first.~~

### DIFF
--- a/pkg/secrethub/http.go
+++ b/pkg/secrethub/http.go
@@ -22,9 +22,10 @@ const (
 	baseURLPath = "/v1"
 
 	// Current account
-	pathMeUser  = "%s/me/user"
-	pathMeRepos = "%s/me/repos"
-	pathMeKey   = "%s/me/key"
+	pathMeUser              = "%s/me/user"
+	pathMeRepos             = "%s/me/repos"
+	pathMeKey               = "%s/me/key"
+	pathMeEmailVerification = "%s/me/user/verification-email"
 
 	// Account
 	pathAccount          = "%s/account/%s"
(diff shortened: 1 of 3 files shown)


pkg/secrethub/http.go
```diff
@@ -153,6 +154,13 @@ func (c *httpClient) GetMyUser() (*api.User, error) {
 	return out, errio.Error(err)
 }
 
+// SendVerificationEmail sends an email to the users registered email address for them to prove they
+// own that email address.
+func (c *httpClient) SendVerificationEmail() error {
+	rawURL := fmt.Sprintf(pathMeEmailVerification, c.base)
+	return c.get(rawURL, nil)
```
mackenbach: This seems kind of weird. Shouldn't it be a `POST` call?

pkg/secrethub/me.go
```diff
@@ -8,15 +8,20 @@ type MeService interface {
...
 }
 
 type meService struct {
+	client      client
 	repoService RepoService
 	userService UserService
 }
 
-func newMeService(repoService RepoService, userService UserService) MeService {
+func newMeService(client client, repoService RepoService, userService UserService) MeService {
```
mackenbach: why add the `repoService` and `userService` as arguments here? What's keeping you from constructing those services inside the `newMeService` function?

SimonBarendse: Because by doing so, the me service only depends on the interfaces, not on the private struct. For now the repo- and userservice can be constructed with only the client as argument, but what if that changes?

One could also argue the other way around. By adding the repo- and userservice as args of newMeService, we expose that the meService uses these internally.

I'm fine with both options.

SimonBarendse: Which do you prefer?

pkg/secrethub/http.go
SimonBarendse: Oh wow, yes this should be a POST call.

pkg/secrethub/me.go
mackenbach: The struct field is the interface, the only surface area where the concrete implrementations touch would be inside the `newMeService` function. Seeing as the alternative is that the exact same surface area would exist in the `client.Me()` function, that's fine. 

When you want to create a custom `MeService` with internal interfaces, you can just create the struct yourself directly.

PR #68, opened 2019-03-28
Add a version badge

SimonBarendse:


florisvdg:
Maybe `release` is nicer than `tag`?



vs.

PR #80, opened 2019-05-30
Decouple credentials

SimonBarendse:
This PR tries to decouple some features of (RSA) credentials. This work is intended to make it possible to add other methods of:
- HTTP Request signing
- Decrypting account keys
- Encrypting account keys


internals/auth/signature.go
```diff
@@ -62,19 +62,31 @@ type Credential interface {
...
+	SignMethod() string
 }
 
-// NewRSACredential initializes a new signing credentials struct.
-func NewRSACredential(key crypto.RSAPrivateKey) Credential {
-	return signer{
-		key: key,
+// HTTPSigner proofs that an HTTP request is made by the owner of the signer.
+type HTTPSigner interface {
+	Sign(r *http.Request) error
```
mackenbach: What makes you require to call this a `Signer` with a `Sign` function? We used to have something in the `auth.Credential` interface called `AddAuthentication`. Then a token-based credential can also add authentication headers through the same interface. 

What's your view on this?

SimonBarendse: Yes, completely agree. :+1:

Also discussed this with Joris this morning, we want the interface to have generic names (authenticater/authenticate) and the constructor more specific (can stay `NewHTTPSigner`)

internals/auth/signature.go
```diff
@@ -115,22 +130,23 @@ func (c signer) AddAuthentication(r *http.Request) error {
...
+	id, err := s.signer.ID()
 	if err != nil {
 		return errio.Error(err)
 	}
 
 	r.Header.Set("Authorization",
-		fmt.Sprintf("%s %s:%s",
-			MethodTagSignature,
-			fingerprint,
+		fmt.Sprintf("%s %s:%s:%s",
```
jpcoenen: Looking into the [Mozilla Developer Page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization), it seems that the standard is to incorporate everything about the type into the first element of the header. So that would make something like:
```
SH1-HMAC-SHA256 <id>:<signature>
```

What do you think about this?

internals/auth/signature_test.go
```diff
@@ -154,6 +143,92 @@ func TestVerify(t *testing.T) {
...
+			},
+			err: auth.ErrNoAuthHeader,
+		},
+		"no key or token": {
+			headers: map[string]string{
+				"Authorization": "TestAuth",
+				"Date":          time.Now().Format(time.RFC3339),
+			},
+			err: auth.ErrUnsupportedAuthFormat,
+		},
```
jpcoenen: Should we add a check for an expired signature?

pkg/secrethub/client.go
```diff
@@ -19,15 +20,29 @@ type Client interface {
 	Users() UserService
 }
 
+// Decryptor decrypts data, typically an account key.
+type Decryptor interface {
+	// Unwrap decrypts data, typically an account key.
+	Unwrap(ciphertext crypto.CiphertextRSAAES) ([]byte, error)
```
jpcoenen: How does this work when the `Decryptor` does not decrypt `RSAAES` ciphertexts?

pkg/secrethub/client.go
```diff
@@ -19,15 +20,29 @@ type Client interface {
...
+}
+
 type clientAdapter struct {
 	client client
 }
 
 // NewClient creates a new SecretHub client.
 // It overrides the default configuration with the options when given.
-func NewClient(credential Credential, opts *ClientOptions) Client {
+func NewClient(decryptor Decryptor, encryptor Encryptor, signer auth.HTTPSigner, opts *ClientOptions) Client {
```
jpcoenen: How would this work if there is no `Encryptor` available in the client? For example, in the case of a service using KMS, the service does not necessarily have the permission to encrypt using KMS.

Do we provide an interface that returns an error or should we remove the `Encryptor` from this signature?

SimonBarendse: I like the idea of providing an interface that returns an error. When implementing it like that, the client doesn't have to check for existence of an encryptor/decryptor at every usage. It even is unaware that encryptor/decryptors can be omitted at all. The message of the error is determined by the user of the client. This is where we know what message to return. We don't have to check for the type of the returned error by the client.

---

Alternatively we could do:
```
client := secrethub.NewClient(signer).Decryptor(decryptor).Encryptor(encryptor)
```

The `Decryptor` and `Encryptor` funcs set the corresponding fields and return a pointer to the client.
When using a func that requires one of them and they have not been supplied (e.g. because a KMS service does not have encrypt permissions) that func returns an error that is defined by the client.

---

Another alternative would be to combine the two by setting default encrypter/decrypter in the constructor that return an error and adding `Decryptor` and `Encryptor` funcs to override these with implementations that can successfully return.

pkg/secrethub/client.go
SimonBarendse: For now, decrypters only decrypt account keys, which are `RSAAES` ciphertexts.

We can change the interface to accept bytes when we return the account key as bytes.

internals/auth/signature.go
mackenbach: Sounds good to me. The types shouldn't vary too much to require many differentiators. A version tag would be nice though.

pkg/secrethub/client.go
mackenbach: If you're going to go with functions that set the decryptor etc., consider using functional options ( http://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis ), e.g. something like this:

```go
client := secrethub.NewClient(authMethod, secrethub.WithDecryptor(key), secrethub.WithEncryptor(key))

// package secrethub
type Option func(c *client) *client

func WithDecryptor(d Decryptor) Option {
   return func(c *client) *client {
      c.decryptor = d
      return c
   }
}
```

The code is quick, dirty and probably somewhat incorrect but you get the idea.

internals/auth/signature.go
SimonBarendse: I've changed it to use `SH1-<signing method> <id>:<signature>`

mackenbach: Ah, so `SH1` is `SecretHub v1`, got it. At a glance I thought it had something to do with `SHA`

internals/auth/signature_test.go
SimonBarendse: yes

SimonBarendse: done

internals/auth/signature.go
SimonBarendse: I've renamed the `HTTPSigner` interface to `Authenticator` with an `Authenticate` function.

internals/auth/signature.go
jpcoenen: Hmm. That can indeed be confusing. We don't want to let people think we're using SHA1.Should we use `SecretHub-v1-` or `SECRETHUB1-` instead?

pkg/secrethub/client.go
jpcoenen: What we should keep in mind that in practice a user of the client does not want to be concerned with what is used as and "encryptor" and what as a "decryptor". In practice the user provides one "method", e.g. "aws with KMS and STS" or "a credential file", and wants everything to just work. I think _we_ should figure out what consequences this has for the encryptor and decryptor, not the user creating the Client.

SimonBarendse: Good point.

How about supplying an interface to the client that has methods to retrieve the encryptor, decryptor and authenticator? We could return such interfaces from the constructors, e.g. `NewCredential`, `NewAWSCredential` etc.

These can in turn return "fake" encryptors, that always return an error when using them.

For the user, usage can become something like this:

```
credential, err := secrethub.NewCredential("<your credential>", "<passphrase>")
client := secrethub.NewClient(credential, nil)
```
and when using AWS KMS:
```
credential, err := secrethub.NewAWSKMSCredential()
client := secrethub.NewClient(credential, nil)
```

jpcoenen: That sounds nice :+1:

jpcoenen: @SimonBarendse , what about this point?

jpcoenen: We leave this for now and pick it up later.

mackenbach: Great insight @jpcoenen, it's really important to keep it simple for the user. Ideally, creating a `Client` should be a oneliner, including credential stuff. The more advanced use cases should be possible, but oneliners should be the default.

mackenbach:
Guys, a small note on naming: 

- `auth.Authenticator` is ambigious. Does it add authentication information to an HTTP request or does it use authentication information to verify that a user is correctly authenticated?
- `auth.Verifier` is a bit better, and explains the difference with `auth.Authenticator`, but that's only if you know the `Verifier` exists.

Consider naming it something like `AddAuth` and `VerifyAuth` to make the distinction clear. Similarly, you could do:

```golang
package auth

type Method interface {
   AddAuth(r *http.Request)
   VerifyAuth(r *http.Request)
}
```

Which immediately makes the different responsibilities clear. Not ultra important now, but something to consider.

PR #81, opened 2019-05-30
Add complex password requirements for randchar package

mackenbach:
This PR adds the ability to mix complex password requirements, e.g. min 2 numbers and no similar characters (`iIlL1oO0`). 

It supports many different character sets and configuration options to meet even the weirdest of password requirements.

I've taken care not to alter the package signature, so existing code won't break. This PR just adds richer functionality under the hood.

Note that this PR doesnt have immediate priority. It's mostly cool features that I've coded on a few lazy sundays. After merging, this functionality could be turned on in the CLI with a handful of new flags.


pkg/randchar/generator.go
```diff
@@ -1,54 +1,290 @@
...
+	}
+}
+
+// Generate returns a randomly generated slice of characters that meets the requirements of the reader.
+func (r Rand) Generate(n int) ([]byte, error) {
+	if n < r.minLen {
+		return nil, errors.New("n cannot be smaller than the minimum required length of the generator")
+	}
+
+	var result []byte
```
golangcibot: Consider preallocating `result` (from `prealloc`)

pkg/randchar/generator_test.go
```diff
@@ -0,0 +1,430 @@
...
+		}
+	}
+	return n
+}
+
+func BenchmarkGenerate24(b *testing.B) {
+	rand := MustNewRand(Alphanumeric)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		rand.Generate(24)
```
golangcibot: Error return value of `rand.Generate` is not checked (from `errcheck`)

pkg/randchar/generator_test.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+}
+
+func BenchmarkGenerate24WithMinimum(b *testing.B) {
+	rand := MustNewRand(Alphanumeric, Min(2, Numeric))
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		rand.Generate(24)
```
golangcibot: Error return value of `rand.Generate` is not checked (from `errcheck`)

pkg/randchar/generator_test.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+}
+
+func BenchmarkGenerate128(b *testing.B) {
+	rand := MustNewRand(Alphanumeric)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		rand.Generate(128)
```
golangcibot: Error return value of `rand.Generate` is not checked (from `errcheck`)

pkg/randchar/generator_test.go
```diff
@@ -0,0 +1,430 @@
...
+		"empty B": {
+			a:        "abc",
+			b:        "",
+			expected: "abc",
+		},
+		"both empty": {
+			a:        "",
+			b:        "",
+			expected: "",
+		},
```
SimonBarendse: How about a testcase for adding sets that are both not a subset of the other for example: "ab", "bc" => "abc"?
In this case neither of the sets is equal to the expected result of the sum of the sets.

pkg/randchar/generator_test.go
```diff
@@ -0,0 +1,430 @@
...
+}
+
+func TestMultipleMin(t *testing.T) {
+	cases := map[string]struct {
+		base                  Charset
+		minima                []minimum
+		expectedInitError     error
+		n                     int
+		expectedGenerateError error
+	}{}
```
SimonBarendse: No cases; Is this still WIP?

pkg/randchar/generator_test.go
```diff
@@ -0,0 +1,430 @@
...
+	}{
+		"empty": {
+			in:       "",
+			expected: "",
+		},
+		"duplicates": {
+			in:       "aa",
+			expected: "a",
+		},
+		"multiple": {
```
SimonBarendse: What are we testing here?

`NewCharset("abc") == Charset("abc")`?

I think it would make more sense to test the underlying byte array in this unit-test and test the equals thing in the Equals unit test (already exists: https://github.com/secrethub/secrethub-go/pull/81/files#diff-ae889a7c8a8642d31b52d28cf0280c68R341).

pkg/randchar/generator.go
```diff
@@ -1,54 +1,306 @@
+// Package randchar helps to generate passwords by allowing you
```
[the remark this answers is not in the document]
mackenbach: Well any text based secret, but that's usually going to be a password.

pkg/randchar/generator.go
```diff
@@ -1,54 +1,306 @@
...
+type Option func(r Rand) (Rand, error)
+
+// minimum defines a minimum number of characters that must be from a given character set.
+type minimum struct {
+	count   int
+	charset Charset
+}
+
+// Min ensures the generated slice contains at least n characters from the given character set.
+// When multiple Min options are given with the same character set, the biggest minimum takes precedence.
```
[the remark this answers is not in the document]
mackenbach: Great that it feels intuitive to you! 😄 Though that may not always be the case, so let's make this behavior explicit in the comment.

pkg/randchar/generator.go
```diff
@@ -1,54 +1,306 @@
...
+		if !exists {
+			result = append(result, char)
+		}
+	}
+
+	return result
+}
+
+// IsSubset returns true when the character set is a subset of the given set.
+// When both sets are the same it returns true too.
```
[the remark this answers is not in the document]
mackenbach: Yes, you're factually correct. Though IMO it's good to be explicit about these behaviors, because not everyone will have this ready top of mind to expect it to work that way.

pkg/randchar/generator.go
```diff
@@ -1,54 +1,306 @@
...
+
+	return result, nil
+}
+
+// shuffle randomly shuffles elements of a byte slice, using the Durstenfeld shuffle algorithm.
+func shuffle(reader io.Reader, data []byte) error {
+	var randomIndex *big.Int
+	var j int64
+	var err error
+	for i := len(data) - 1; i > 0; i-- {
```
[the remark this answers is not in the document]
mackenbach: Wow that could be dangerous! After some investigation, I think it **does** get swapped. 

For each element `i`, it is swapped with another randomly chosen element `j` where `0 <= j <= i`. This means that on each iteration through the `for` loop, `data[0]` has a chance to be selected as the element to swap with `i`. 

Good call to check it, though I think we're safe 😄 

For some background, check out these resources:

- Algorithm description: https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm
- Also, the `math/rand` package seems to have implemented the Fisher-Yates algorithm (which is similar) for their `Shuffle` function: https://golang.org/src/math/rand/rand.go?s=11285:11325#L232 It's a bit less readable though.

pkg/randchar/generator.go
```diff
@@ -1,54 +1,310 @@
...
+		// Note that the algorithm requires j to be a uniform random
+		// value such that 0 <= j <= i and rand.Int returns a value
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

pkg/randchar/generator.go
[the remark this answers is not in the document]
mackenbach: Aha, but I did identify a different bug! Turns out `rand.Int(reader, i)` produces a value `[0, i)`, which is _exclusive_ `i` and we need it to _inclusive_ `i`.

See eb2e164 for the fix.

pkg/randchar/generator_test.go
mackenbach: Was tested elsewhere but forgot to remove this test function.

pkg/randchar/generator.go
```diff
@@ -1,54 +1,306 @@
...
+		}
+		j = randomIndex.Int64()
+		data[i], data[j] = data[j], data[i]
+	}
+
+	return nil
+}
+
+// Charset is a byte slice with a set of unique characters.
+type Charset []byte
```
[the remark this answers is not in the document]
mackenbach: Yes we could, though it's a minor optimization for something that is called almost never (a few times on startup). Moreover, it keeps the logic for selecting a random character from the set more straightforward. Instead of selecting a random index from `map` keys (which has some undeterministic logic behind it) we select a random index from a fixed byte slice which is deterministic. In other words, it has less moving parts.

We can optimize the internal representation of the character set later IMO. Not something we'd need to dwell on now. We could make it a private type and only expose the `NewCharset` operation, but then we'd have to return an interface.

mackenbach:
I've implemented most of your remarks and replied to the rest. @SimonBarendse could you re-review?

mackenbach:
> Before adding this to the CLI, let's discuss (in person?) whether this functionality falls within its scope. Similar functionality could be achieved by piping the output of an external tool to secrethub write.

I do feel this should be included in the CLI. We want the CLI to be used to generate passwords/secrets for real applications and sadly real applications sometimes impose weird password restrictions. When that happens, we want to be flexible enough to help users satisfy those requirements. Using a second tool to generate the secret and piping it to `write` detracts from the power of the CLI. 

Moreover, we don't only want to offer this flexible functionality through the CLI, but also through the Terraform provider, Ansible module and other integrations. Imagine having to use a different tool to generate the value and then passing it to `secrethub_secret` as opposed to adding `min` option to the `secrethub_secret` resource.

pkg/randchar/generator.go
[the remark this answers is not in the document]
SimonBarendse: Ahh, I was thinking because `i` never reached 0, data[0] would never get swapped. However, as you point out `j` can reach 0. So data[0] can get swapped. :+1: 

`i > 0` is the correct boundary, as having `i=0` would make `j=0` (random integer in the [0,0] range). This wouldn't change anything, as you'd swap `data[0]` with `data[0]`.

pkg/randchar/generator.go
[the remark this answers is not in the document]
SimonBarendse: I like your proposal of making the charset type private and exposing an interface. This makes the internal representation unexposed, which makes it possible to change it later without breaking changes. :+1: 

Note that we can't make any guarantees of the charset operations only being used on startup. They are public operations in a public package, so they can be used in any way.

I agree that the charsets should be optimized for selecting random characters. That's what they are meant to be used for.
What do you mean with undeterministic logic? Isn't (selecting) something random supposed to be undeterministic?

pkg/randchar/generator_test.go
```diff
@@ -175,6 +175,16 @@ func TestAdd(t *testing.T) {
 			b:        "",
 			expected: "",
 		},
+		"different": {
+			a:        "ab",
+			b:        "bc",
+			expected: "abc",
+		},
+		"different with overlap": {
```
SimonBarendse: I think the names of these cases are swapped. `"ab", "bc"` have overlap, `"ab", "c"` have not.

pkg/randchar/generator.go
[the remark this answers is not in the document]
mackenbach: This is more gut feeling than anything else, but I like to know where my randomness comes from. A loop over a slice is fixed and randomness comes from `crypto/rand`. A loop over map keys is performed in random order, which adds additional randomness to that coming from `crypto/rand`. That's why I prefer the slice type, you know what you're dealing with. 

You're right about us not being able to give guarantees about where it's used. This is meant as a library so creating an interface would actually help to minimize the exposure and keep internal flexibility for later. Will add that when I have time.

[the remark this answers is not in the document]
SimonBarendse: Let's keep it a byte slice, but hide the internal type by making it private and returning an interface, like you proposed. :+1:

pkg/randchar/generator_test.go
mackenbach: We're testing whether multiple characters leads to a set with multiple characters. It's not a super interesting test, but we should have it nonetheless IMO. I changed it so the equality test is indeed more a unit test and doesn't depend on the `Charset.Equals` test.

mackenbach:
Isolated the `Charset` type and implemented your other comments. Turned out that creating an interface was way overkill for this and resulted in type casts when you want to compare one set to another etc. So I ended up making it a pubic type with a private field so we have the protection from (unintentional) abuse of the `Charset` type without the hassle of an interface.

@SimonBarendse could you check it out?

pkg/randchar/generator.go
```diff
@@ -251,36 +255,48 @@ func (cs Charset) Add(set Charset) Charset {
...
 
 // IsSubset returns true when the character set is a subset of the given set.
 // When both sets are the same it returns true too.
-func (cs Charset) IsSubset(of Charset) bool {
+func (c Charset) IsSubset(of Charset) bool {
+	if c.Size() == 0 {
+		return true
+	}
+
+	if c.Size() > 0 && of.Size() == 0 {
```
SimonBarendse: These size checks are redundant, why would we add them?

pkg/randchar/generator.go
```diff
@@ -213,8 +213,10 @@ func shuffle(reader io.Reader, data []byte) error {
 	return nil
 }
 
-// Charset is a byte slice with a set of unique characters.
-type Charset []byte
+// Charset is a set of unique characters.
+type Charset struct {
+	chars []byte
```
SimonBarendse: :ok_hand:

pkg/randchar/generator.go
SimonBarendse: I mean that

```Go
if c.Size() == 0 {
	return true
}

if c.Size() > 0 && of.Size() == 0 {
	return false
}
```

can be removed, as it is also covered by the set logic below.

mackenbach: Yes, functionally they are the same but this adds a small speed optimization. It's not super big, but we can leave it in. Clarified it by restructuring the `if` statements and with a comment.

PR #86, opened 2019-06-07
Accept a string instead of api.Permission on AccessRule.Set

SimonBarendse:


SimonBarendse:
@mackenbach 

> It looks okay to me, but I don't immediately see why we would want to change this. Could you clarify your motivation for this PR?

In https://github.com/secrethub/secrethub-go/pull/28 we've changed the client signatures to accept primitive types instead of api types for a.o. secret paths. We've done so because we want the user to be able to just supply the values without having to validate and convert to some internal type. Instead, we validate and convert inside the client.

Similarly, we don't want the user to be bothered with validating and converting permission strings (read, write, admin etc.).

Instead of

```go
var perm api.Permission
err := perm.Set(permission)
if err != nil {
	return nil, err
}

client.AccessRules().Set(path, perm, account)
```

the user of the client can just do

```go
client.AccessRules().Set(path, permission, account)
```

PR #90, opened 2019-06-20
Return RSACredential instead of Credential

jpcoenen:
Since we are handling encoding and decoding here, passing around an interface is a pain. By saying these functions can only return RSACredential's, we remove a lot of the complexity. This is needed to make the CLI upgrade to the new version of secrethub-go a lot smoother.

In combination with https://github.com/secrethub/secrethub-cli/pull/104.

Tests still have to be fixed.

### DIFF
--- a/pkg/secrethub/credential.go
+++ b/pkg/secrethub/credential.go
@@ -94,7 +94,7 @@ func NewCredential(credential string, passphrase string) (Credential, error) {
 // CredentialDecoder converts a payload into a Credential.
 type CredentialDecoder interface {
 	// Decode decodes a payload into a Credential.
-	Decode(payload []byte) (Credential, error)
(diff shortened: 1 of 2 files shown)


pkg/secrethub/credential.go
```diff
@@ -275,17 +275,17 @@ type RSACredential struct {
...
-func generateRSACredential(keyLength int) (RSACredential, error) {
+func generateRSACredential(keyLength int) (*RSACredential, error) {
 	key, err := crypto.GenerateRSAPrivateKey(keyLength)
 	if err != nil {
-		return RSACredential{}, errio.Error(err)
+		return nil, errio.Error(err)
 	}
 
-	return RSACredential{
+	return &RSACredential{
```
SimonBarendse: Why are we now returning pointers here?

PR #91, opened 2019-07-01
AWS Integration

jpcoenen:
Allow the creation of and usage of services that use AWS STS for authentication and AWS KMS for Account Key decryption.

There is a lot in this PR. Therefore I have tried to describe individual parts with links to the code below. Overall, the best method to review is to do a compare directory by directory against `develop`, using the descriptions below as a guideline.

1. Add [`api.EncryptedData`](https://github.com/secrethub/secrethub-go/blob/feature/aws-integration-merge/internals/api/encrypted_data.go) as a new type to store/transfer generic encrypted data. In this PR it is only implemented with Encrypted Account Keys. Old clients will get the old `api.CiphertextRSAAES` response. See https://github.com/secrethub/secrethub-go/pull/89 for more details.
2. Add the creation of services that use AWS STS for authentication and AWS KMS for key encryption (see Figure 1):
     - [Add](https://github.com/secrethub/secrethub-go/pull/91/files#diff-bb20642f1ccdcf84d5724fd96209508cR60) the `Proof` field to `api.CreateCredentialRequest`. This is used by the server to verify the possession of the credential. For the case of AWS-STS this is done by adding a separate AWS KMS Encrypt request containing a known plaintext that is sent to the AWS by the API Server (to verify its validity).
      - [ `aws.ServiceCreator`](https://github.com/secrethub/secrethub-go/blob/feature/aws-integration-merge/internals/aws/service_creator.go) is a struct the implements the `secrethub.Encrypter` and `secrethub.Verifier` interfaces to correctly create a credential and encrypted account key for a service using AWS.
      - Add [`secrethub.ServiceAWSService`](https://github.com/secrethub/secrethub-go/blob/feature/aws-integration-merge/pkg/secrethub/service_aws.go) to handle creation of AWS services. This is a sub-service of the `secrethub.ServiceService` service.

image
_Figure 1: Creating AWS Service_

3. Add [secrethub.AuthService](https://github.com/secrethub/secrethub-go/blob/feature/aws-integration-merge/pkg/secrethub/auth.go), corresponding [`secrethub.awsAuthService`](https://github.com/secrethub/secrethub-go/blob/feature/aws-integration-merge/pkg/secrethub/auth_aws.go) and [API types](https://github.com/secrethub/secrethub-go/blob/feature/aws-integration-merge/internals/api/auth.go) to handle authentication to the server using AWS STS. This creates a AWS STS `GetCallerIdentity` request without sending it to AWS. This request is then sent to the SecretHub API to verify the caller's identity and lookup any corresponding service accounts. The server returns an [`api.Session`](https://github.com/secrethub/secrethub-go/pull/91/files#diff-1e3b9e873de9f6a4f9b657370848ab54R150)(new type). See Figure 2 for an overview of this process.

image
_Figure 2: Authentication using AWS STS_

4. Add authentication using HMAC sessions  by implementing [`auth.SessionSigner`](https://github.com/secrethub/secrethub-go/blob/feature/aws-integration-merge/internals/auth/session.go) that takes a valid session and uses it to sign requests to the API.

5. Add a [`aws.KMSDecrypter`](https://github.com/secrethub/secrethub-go/blob/feature/aws-integration-merge/internals/aws/kms_decrypter.go) that implements the `secrethub.Decrypter` interface and uses AWS KMS to decrypt an encrypted AccountKey.

5. Add [`secrethub.NewClientAWS()`](https://github.com/secrethub/secrethub-go/pull/91/files#diff-d70e32369866756c84888f2b28e19b66R49). This creates a new client with `aws.KMSDecrypter` as decrypter and [`auth.NopAuthenticator`](https://github.com/secrethub/secrethub-go/blob/feature/aws-integration-merge/internals/auth/nop.go) (a new authenticator that does not add any authentication) as authenticator. Then it calls `client.Auth().AWS().Authenticate()` to get authenticate using AWS STS and then get an authenticator that uses a session. The `NopAuthenticator` is then replaced by this authenticator and the client is good to go.

This change set is not well tested yet. That is work-in-progress.

_Note: figure sources are in the source of this PR below this note._


pkg/secrethub/fakeclient/service.go
```diff
@@ -9,14 +9,15 @@ import (
 
 // ServiceService is a mock of the ServiceService interface.
 type ServiceService struct {
-	Creater ServiceCreater
-	Deleter ServiceDeleter
-	Getter  ServiceGetter
-	Lister  RepoServiceLister
+	Creater    ServiceCreater
```
golangcibot: `Creater` is a misspelling of `Creature` (from `misspell`)

pkg/secrethub/fakeclient/service_aws.go
```diff
@@ -0,0 +1,39 @@
...
+package fakeclient
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/secrethub/secrethub-go/internals/api"
+)
+
+// ServiceAWSService is a mock of the ServiceAWSService interface.
+type ServiceAWSService struct {
+	Creater ServiceAWSCreater
```
golangcibot: `Creater` is a misspelling of `Creature` (from `misspell`)

internals/aws/service_creator.go
```diff
@@ -0,0 +1,130 @@
...
+
+// NewServiceCreator returns a ServiceCreator that uses the provided AWS KMS key and IAM role to create a new service.
+// The AWS service is configured with the optionally provided aws.Config.
+func NewServiceCreator(keyID, role string, cfgs ...*aws.Config) (*ServiceCreator, error) {
+	sess, err := session.NewSession(cfgs...)
+	if err != nil {
+		return nil, handleError(err)
+	}
+
+	getAccountId := func() (string, error) {
```
golangcibot: var `getAccountId` should be `getAccountID` (from `golint`)

internals/aws/service_creator_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/aws/aws-sdk-go/service/kms/kmsiface"
+
+	"github.com/aws/aws-sdk-go/service/sts"
+
+	"github.com/aws/aws-sdk-go/service/sts/stsiface"
+
+	"github.com/secrethub/secrethub-go/internals/assert"
+)
+
+var defaultTestErr = errors.New("test-error")
```
golangcibot: error var defaultTestErr should have name of the form errFoo (from `golint`)

pkg/secrethub/auth.go
```diff
@@ -0,0 +1,28 @@
...
+import "github.com/secrethub/secrethub-go/internals/auth"
+
+// AuthMethodService is an interface for any service that can provide authentication to the server.
+type AuthMethodService interface {
+	Authenticate() (auth.Authenticator, error)
+}
+
+// AuthService handles authentication to the SercretHub API.
+type AuthService interface {
+	AWS() AuthMethodService
```
mackenbach: Why does this service have only one function? Should it also have `RSASigner` or `Session` or something like that? This suggests there is only one way to authenticate.

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+
+func newAWSAuthService(client client) AuthMethodService {
+	return &awsAuthService{
+		client: client,
+	}
+}
+
+func (s awsAuthService) Authenticate() (auth.Authenticator, error) {
+	// Currently always use the eu-west-1 region.
+	region := endpoints.EuWest1RegionID
```
mackenbach: Can we make this a constant and maybe place it at the top of the file, calling it `defaultAWSRegionForSTS` or something like that?

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+
+	cfg := aws.NewConfig().WithRegion(region).WithEndpoint("sts." + region + ".amazonaws.com")
```
mackenbach: This looks kinda weird: you're passing region to it and then `WithEndpoint` in which you have to set the region again. If this is awkwardness is AWS's design then cool, if it's because we're using it wrong let's see how we can improve. 

I like their use of functional options though 😍 (although it could have even been better if it wasn't used on configs but on the actual client)

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+	awsSess, err := session.NewSession(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("could not get AWS session: %v", err)
+	}
+
+	svc := sts.New(awsSess, cfg)
```
mackenbach: Please no more `svc` 😢

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+	client client
+}
+
+func newAWSAuthService(client client) AuthMethodService {
+	return &awsAuthService{
+		client: client,
+	}
+}
+
+func (s awsAuthService) Authenticate() (auth.Authenticator, error) {
```
mackenbach: Add godoc comments por favor

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+
+	cfg := aws.NewConfig().WithRegion(region).WithEndpoint("sts." + region + ".amazonaws.com")
+	awsSess, err := session.NewSession(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("could not get AWS session: %v", err)
+	}
+
+	svc := sts.New(awsSess, cfg)
+
+	stsReq, _ := svc.GetCallerIdentityRequest(&sts.GetCallerIdentityInput{})
```
mackenbach: Consider naming this something a bit more verbose like `identityRequest`

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+
+	err = stsReq.Sign()
```
mackenbach: `// Sign the CallerIdentityRequest with the AWS access key`

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil, fmt.Errorf("could not sign STS request: %v", err)
+	}
+
+	var buf bytes.Buffer
+	err = stsReq.HTTPRequest.Write(&buf)
+	if err != nil {
+		return nil, err
+	}
+
+	req := api.NewAuthRequestAWSSTS(api.SessionTypeHMAC, region, buf.Bytes())
```
mackenbach: This raises the question: where is HMAC used? I see now it's the returned authentication type thingy.

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+	err = stsReq.Sign()
+	if err != nil {
+		return nil, fmt.Errorf("could not sign STS request: %v", err)
+	}
+
+	var buf bytes.Buffer
+	err = stsReq.HTTPRequest.Write(&buf)
+	if err != nil {
+		return nil, err
+	}
```
mackenbach: Consider moving this sts prepare request logic  to a separate helper function to auto document the behavior.

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+
+	req := api.NewAuthRequestAWSSTS(api.SessionTypeHMAC, region, buf.Bytes())
+	resp, err := s.client.httpClient.Authenticate(req)
```
mackenbach: Should this be `GetSession` or `CreateSession`?

pkg/secrethub/auth_aws.go
mackenbach: I feel we need to look at the contract that this exposes. In my mind, I should be able to say something like:

```go
client := secrethub.NewClient()
client := secrethub.NewClient().WithAWS(foo)?
client := secrethub.NewClientWithAWS(foo)?
```

mackenbach: Consider making it private or creating a distinction between plain clients and authenticated clients, e.g. 

```go
func (c client) CreateSession(foo) Client {
...
}
```

pkg/secrethub/client.go
```diff
@@ -46,6 +46,23 @@ func NewClient(decrypter Decrypter, authenticator auth.Authenticator, opts *Clie
...
+// NewClientAWS creates a new SecretHub client that uses AWS STS and KMS to access SecretHub.
+func NewClientAWS(opts *ClientOptions) (Client, error) {
+	decrypter, err := aws.NewKMSDecrypter()
+	if err != nil {
+		return nil, err
+	}
+	client := &clientAdapter{
+		client: newClient(decrypter, auth.NopAuthenticator{}, opts),
+	}
+	authenticator, err := client.Auth().AWS().Authenticate()
```
mackenbach: Structural comment: this is exposed to users of the library. Do we want that, or do we want to keep this stuff private and only expose higher level functions such as `NewClientAWS`?

internals/auth/nop.go
```diff
@@ -0,0 +1,11 @@
+package auth
+
+import "net/http"
+
+// NopAuthenticator is an authenticator that does not add any authentication to the request.
+type NopAuthenticator struct{}
```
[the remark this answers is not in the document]
mackenbach: Also, it's public. Should it be?

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if *resp.Type != api.SessionTypeHMAC {
+		return nil, api.ErrInvalidSessionType
+	}
+	sess := resp.HMAC()
```
mackenbach: Can we make this `session` or something? This is confusing. Overall, let's try to make this file a bit clearer with more helper functions, comments and clearer naming.

pkg/secrethub/auth_aws.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return auth.NewHTTPSigner(auth.NewSessionSigner(sess.SessionID, api.StringValue(sess.Payload.SecretKey))), nil
```
mackenbach: Why does this use a `SecretKey`? I thought secret keys were used to encrypt secrets?

pkg/secrethub/client.go
```diff
@@ -11,6 +12,7 @@ import (
 type Client interface {
 	AccessRules() AccessRuleService
 	Accounts() AccountService
+	Auth() AuthService
```
mackenbach: Is it `Auth()` or `Sessions()`?

pkg/secrethub/client.go
```diff
@@ -46,6 +46,23 @@ func NewClient(decrypter Decrypter, authenticator auth.Authenticator, opts *Clie
...
 }
 
+// NewClientAWS creates a new SecretHub client that uses AWS STS and KMS to access SecretHub.
+func NewClientAWS(opts *ClientOptions) (Client, error) {
+	decrypter, err := aws.NewKMSDecrypter()
+	if err != nil {
+		return nil, err
+	}
+	client := &clientAdapter{
+		client: newClient(decrypter, auth.NopAuthenticator{}, opts),
```
mackenbach: I think we should look into decorating the client with authentication instead of injecting authentication into the client. That's causing us to need a `NopAuthenticator`.

pkg/secrethub/client.go
```diff
@@ -46,6 +46,23 @@ func NewClient(decrypter Decrypter, authenticator auth.Authenticator, opts *Clie
...
+	}
+	authenticator, err := client.Auth().AWS().Authenticate()
+	if err != nil {
+		return nil, err
+	}
+	client.client.httpClient.authenticator = authenticator
```
mackenbach: Haha this is a weird assignment. Necessary though because of the injection thing.

pkg/secrethub/credential.go
```diff
@@ -94,7 +94,7 @@ func NewCredential(credential string, passphrase string) (Credential, error) {
 // CredentialDecoder converts a payload into a Credential.
 type CredentialDecoder interface {
 	// Decode decodes a payload into a Credential.
-	Decode(payload []byte) (Credential, error)
+	Decode(payload []byte) (*RSACredential, error)
```
mackenbach: If this is all RSA specific, should we move it to a separate file or rename this one?

pkg/secrethub/credential.go
```diff
@@ -321,31 +325,77 @@ func (c RSACredential) Decoder() CredentialDecoder {
...
 // Wrap encrypts data, typically an account key.
-func (c RSACredential) Wrap(plaintext []byte) (crypto.CiphertextRSAAES, error) {
-	return c.RSAPrivateKey.Public().Encrypt(plaintext)
+func (c RSACredential) Wrap(plaintext []byte) (*api.EncryptedData, error) {
+	ciphertext, err := c.RSAPrivateKey.Public().Encrypt(plaintext)
+	if err != nil {
+		return nil, err
+	}
+	key := api.NewEncryptionKeyEncrypted(
+		crypto.SymmetricKeyLength*8,
```
mackenbach: This looks pretty specific to do here. Can we create a function `ciphertext.ToEncryptedData()` instead, e.g. in the `crypto` package?

pkg/secrethub/credential.go
```diff
@@ -321,31 +325,77 @@ func (c RSACredential) Decoder() CredentialDecoder {
...
+		return nil, api.ErrInvalidCiphertext
+	}
+	return c.RSAPrivateKey.Decrypt(crypto.CiphertextRSAAES{
+		AES: crypto.CiphertextAES{
+			Data:  ciphertext.Ciphertext,
+			Nonce: metadata.Nonce,
+		},
+		RSA: crypto.CiphertextRSA{
+			Data: encryptedKey.Ciphertext,
+		},
```
mackenbach: In the future, will we even need this `crypto.CiphertextRSAAES` type? I feel like the `EncryptedData` type could cover it pretty nicely.

pkg/secrethub/fakeclient/service.go
mackenbach: Is it `Creator` or `Creater` or `CreateFunc`?

pkg/secrethub/http.go
```diff
@@ -21,10 +21,12 @@ var (
 const (
 	baseURLPath = "/v1"
 
+	pathAuthenticate = "%s/auth"
+
 	// Current account
 	pathMeUser              = "%s/me/user"
 	pathMeRepos             = "%s/me/repos"
-	pathMeKey               = "%s/me/key"
+	pathMeKey               = "%s/me/key?key_version=v2"
```
mackenbach: I'd stay away from manipulating the path here. Not 100% sure, but it does assume this path is an absolute path and cannot be joined with anything else. Although, the other paths are that as well. Think about it and maybe just ignore my ramblings here.

pkg/secrethub/http.go
```diff
@@ -121,6 +123,13 @@ func newHTTPClient(authenticator auth.Authenticator, opts *ClientOptions) *httpC
 	}
 }
 
+func (c *httpClient) Authenticate(in interface{}) (*api.Session, error) {
```
mackenbach: As said before, this is far more a `CreateSession` than `Authenticate` which is an ambiguous word to me.

pkg/secrethub/http.go
```diff
@@ -121,6 +123,13 @@ func newHTTPClient(authenticator auth.Authenticator, opts *ClientOptions) *httpC
...
+	var out api.Session
+	rawURL := fmt.Sprintf(pathAuthenticate, c.base)
+	err := c.post(rawURL, http.StatusOK, in, &out)
```
mackenbach: Should this expect `http.StatusOK` or `http.StatusCreated`?

pkg/secrethub/service.go
```diff
@@ -8,13 +8,15 @@ import (
 // ServiceService handles operations on service accounts from SecretHub.
 type ServiceService interface {
 	// Create creates a new service account for the given repo.
-	Create(path string, description string, credential Credential, encrypter Encrypter) (*api.Service, error)
+	Create(path string, description string, verifier Verifier, encrypter Encrypter) (*api.Service, error)
```
mackenbach: In practice, what does a call to this function look like?

```go
// RSA
client.Services().Create("company/repo", "my service", credential, credential)
// AWS
client.Services().AWS().Create(path, description, keyID, role, cfgs...)
```

You already have a separate function for AWS, so for the RSA function why not combine `verifier` and `encrypter` in the same argument with a joined interface?

mackenbach: Also, how do you feel about having the interface composed this way with a separate AWS interface but having the `Create` function on the 'root' `ServiceService` interface?

pkg/secrethub/service_aws.go
```diff
@@ -0,0 +1,35 @@
...
+	serviceService ServiceService
+}
+
+// Create creates a new AWS service account for the given repo.
+func (s serviceAWSService) Create(path string, description string, keyID, role string, cfgs ...*awssdk.Config) (*api.Service, error) {
+	creator, err := aws.NewServiceCreator(keyID, role, cfgs...)
+	if err != nil {
+		return nil, err
+	}
+	return s.serviceService.Create(path, description, creator, creator)
```
mackenbach: Huh, interesting... You're calling the `Create` again here. I'm starting to wonder whether the `create` with separate `verifier` and `encrypter` arguments should be private and we expose only public functions that are specific to the type of credential?

pkg/secrethub/credential.go
```diff
@@ -39,12 +39,12 @@ var (
...
 // Verifier exports verification bytes that can be used to verify signed data is processed by the owner of a signer.
 type Verifier interface {
-	// Fingerprint returns an identifier by which the server can identify the credential, e.g. a username of a fingerprint.
-	Fingerprint() (string, error)
 	// Verifier returns the data to be stored server side to verify an http request authenticated with this credential.
 	Verifier() ([]byte, error)
 	// Type returns what type of credential this is.
 	Type() api.CredentialType
+	// AddProof adds the proof of this credential's possession to a CreateCredentialRequest.
+	AddProof(req *api.CreateCredentialRequest) error
```
mackenbach: Is this every used? Cannot find where it is called.

pkg/secrethub/fakeclient/client.go
```diff
@@ -24,6 +25,11 @@ func (c Client) Accounts() secrethub.AccountService {
 	return c.AccountService
 }
 
+// Auth implements the secrethub.Client interface.
+func (c Client) Auth() secrethub.AuthService {
+	return c.AuthService
```
golangcibot: cannot use c.AuthService (variable of type *AuthService) as secrethub.AuthService value in return statement: wrong type for method AWS (from `typecheck`)

pkg/secrethub/fakeclient/service.go
jpcoenen: `Creator` might be better, as `Creater` is incorrect English. And convention is to use correct English over `+er`.

internals/api/auth.go
```diff
@@ -0,0 +1,249 @@
...
+	SessionTypeHMAC SessionType = "hmac"
+)
+
+// Errors
+var (
+	ErrInvalidSessionType  = errAPI.Code("invalid_session_type").StatusError("invalid session type provided for authentication request", http.StatusBadRequest)
+	ErrInvalidPayload      = errAPI.Code("invalid_payload").StatusError("invalid payload provided for authentication request", http.StatusBadRequest)
+	ErrInvalidAuthMethod   = errAPI.Code("invalid_auth_method").StatusError("invalid auth method", http.StatusBadRequest)
+	ErrMissingField        = errAPI.Code("missing_field").StatusErrorPref("request is missing field %s", http.StatusBadRequest)
+	ErrSessionNotFound     = errAPI.Code("session_not_found").StatusError("session could not be found, it might have expired", http.StatusForbidden)
```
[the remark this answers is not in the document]
jpcoenen: We don't know whether it does not exist or the TTL in Redis has expired. It is _gone_.

internals/api/auth.go
```diff
@@ -0,0 +1,249 @@
...
+	ErrSessionExpired      = errAPI.Code("session_expired").StatusError("session has expired", http.StatusForbidden)
+	ErrAuthFailed          = errAPI.Code("auth_failed").StatusError("authentication failed", http.StatusForbidden)
+	ErrCouldNotGetEndpoint = errAPI.Code("wrong_endpoint").StatusError("could not find the AWS endpoint", http.StatusBadRequest)
```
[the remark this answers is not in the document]
jpcoenen: We're talking about the AWS endpoint the server has to contact to pass on a request (either `GetCallerIdentity` or `Encrypt`). The only legit reason why a user would see this is if the provided AWS region is incorrect.

internals/api/auth.go
```diff
@@ -0,0 +1,249 @@
...
+	ErrAWSException        = errAPI.Code("aws_exception").StatusError("AWS returned an error", http.StatusFailedDependency)
```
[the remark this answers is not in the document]
jpcoenen: I understand your concerns and I agree this can be improved. This is returned if an AWS response we get on the server is unexpected. I don't think we want to expose any of this to the user, so the content of the error message should be vague (like an internal server error; though this one is not inernal). However, we could make the message a bit more descriptive like the error message for unexpected errors.

internals/api/auth.go
```diff
@@ -0,0 +1,249 @@
...
+
+// Errors
+var (
+	ErrInvalidSessionType  = errAPI.Code("invalid_session_type").StatusError("invalid session type provided for authentication request", http.StatusBadRequest)
+	ErrInvalidPayload      = errAPI.Code("invalid_payload").StatusError("invalid payload provided for authentication request", http.StatusBadRequest)
+	ErrInvalidAuthMethod   = errAPI.Code("invalid_auth_method").StatusError("invalid auth method", http.StatusBadRequest)
+	ErrMissingField        = errAPI.Code("missing_field").StatusErrorPref("request is missing field %s", http.StatusBadRequest)
+	ErrSessionNotFound     = errAPI.Code("session_not_found").StatusError("session could not be found, it might have expired", http.StatusForbidden)
+	ErrSessionExpired      = errAPI.Code("session_expired").StatusError("session has expired", http.StatusForbidden)
+	ErrAuthFailed          = errAPI.Code("auth_failed").StatusError("authentication failed", http.StatusForbidden)
```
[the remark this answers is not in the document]
jpcoenen: This is returned if AWS returns a `http.StatusForbidden`. I haven't found a case where this happens. I first thought this would be if authentiction is incorrect, but then AWS returns a BadRequest. 

So we could opt to drop this.

internals/api/auth.go
```diff
@@ -0,0 +1,249 @@
...
+	ErrNoServiceWithRole   = errAPI.Code("no_service_with_role").StatusErrorPref("there exists no service account tied to the AWS IAM role '%s'", http.StatusNotFound)
+)
+
+// SessionType defines how a session can be used.
+type SessionType string
+
+// AuthRequest is a request to authenticate and request a session.
+type AuthRequest struct {
+	Method      *string      `json:"method"`
+	SessionType *SessionType `json:"session_type"`
```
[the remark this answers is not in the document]
jpcoenen: For unmarshaling it makes sense. This way you can detect if a field is not set instead of empty. `omitempty` has no effect on this. So if someone uses our API without using our client, they get more usable error messages.

internals/api/auth.go
```diff
@@ -0,0 +1,249 @@
...
+	return nil
+}
+
+// NewSessionHMAC returns a HMAC type api.Session.
+func NewSessionHMAC(sessionID uuid.UUID, expiration time.Time, secretKey string) *Session {
+	t := SessionTypeHMAC
+	return &Session{
+		SessionID:  &sessionID,
+		Expiration: &expiration,
+		Type:       &t,
```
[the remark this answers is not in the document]
jpcoenen: Yeah. But I don't see a reason for using them if your value already is in a variable. Then `&` work just as well.

internals/api/auth.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// SessionPayloadHMAC is the payload of a HMAC typed session.
+type SessionPayloadHMAC struct {
+	SecretKey *string `json:"secret_key"`
+}
+
+// SessionHMAC is an HMAC-specific representation of a session.
+type SessionHMAC struct {
+	SessionID  uuid.UUID
```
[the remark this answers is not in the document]
jpcoenen: For know these are never marshaled nor unmarshaled to JSON directly, that's why they're missing. If we see a reason for calling `json` functins on this struct, we should add them.

internals/api/auth.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+type validator interface {
+	Validate() error
+}
+
+// Validate whether the Session is valid.
+func (s *Session) Validate() error {
+	if s.SessionID == nil {
+		return ErrMissingField("session_id")
```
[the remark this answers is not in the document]
jpcoenen: You mean `session_id: ""`? I think that falls in the same category as `session_id: "a"`.

We can opt to actually validate whether it is a value UUID, but discerning between a value set as empty and an illegal value does not add any value if you ask me.

internals/api/credential.go
```diff
@@ -9,9 +14,14 @@ import (
 
 // Errors
 var (
-	ErrInvalidFingerprint = errAPI.Code("invalid_fingerprint").StatusError("fingerprint is invalid", http.StatusBadRequest)
-	ErrInvalidVerifier    = errAPI.Code("invalid_verifier").StatusError("verifier is invalid", http.StatusBadRequest)
-	ErrInvalidAlgorithm   = errAPI.Code("invalid_algorithm").StatusError("algorithm is invalid", http.StatusBadRequest)
+	ErrInvalidFingerprint    = errAPI.Code("invalid_fingerprint").StatusError("fingerprint is invalid", http.StatusBadRequest)
```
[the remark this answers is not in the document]
jpcoenen: It's a credential fingerprint. It's an error a normal user should not see as the generation of the fingerprint is handled by the client.

internals/api/credential.go
```diff
@@ -9,9 +14,14 @@ import (
...
+	ErrInvalidVerifier       = errAPI.Code("invalid_verifier").StatusError("verifier is invalid", http.StatusBadRequest)
+	ErrInvalidCredentialType = errAPI.Code("invalid_credential_type").StatusError("credential type is invalid", http.StatusBadRequest)
+	ErrInvalidAWSEndpoint    = errAPI.Code("invalid_aws_endpoint").StatusError("invalid AWS endpoint provided", http.StatusBadRequest)
+	ErrInvalidProof          = errAPI.Code("invalid_proof").StatusError("invalid proof provided for credential", http.StatusBadRequest)
```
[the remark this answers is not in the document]
jpcoenen: `http.StatusUnauthorized` makes sense :+1:

internals/api/credential.go
```diff
@@ -9,9 +14,14 @@ import (
...
+	ErrAWSAccountMismatch    = errAPI.Code("aws_account_mismatch").StatusError("role account id does not match with authentication account id. Make sure you are using AWS credentials that correspond to the role you are trying to add.", http.StatusBadRequest)
```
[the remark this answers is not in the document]
jpcoenen: That's exactly what I mean

internals/api/credential.go
```diff
@@ -9,9 +14,14 @@ import (
...
+	ErrAWSAuthFailed         = errAPI.Code("aws_auth_failed").StatusError("authentication not accepted by AWS", http.StatusForbidden)
```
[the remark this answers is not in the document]
jpcoenen: See previous discussion about this.

internals/api/credential.go
```diff
@@ -9,9 +14,14 @@ import (
...
+	ErrInvalidRoleARN        = errAPI.Code("invalid_role_arn").StatusError("provided role is not a valid ARN", http.StatusBadRequest)
```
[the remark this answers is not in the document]
jpcoenen: As far as I know, it's AWS specific.

internals/api/credential.go
```diff
@@ -29,39 +39,132 @@ type CredentialType string
...
+		}
+	}
+	*req = CreateCredentialRequest(dec)
+	return nil
+}
 
+// Validate validates the request fields.
+func (req *CreateCredentialRequest) Validate() error {
+	if req.Fingerprint == nil {
+		return ErrMissingField("fingerprint")
```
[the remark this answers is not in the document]
jpcoenen: Then the check a few lines below fails.

internals/api/credential.go
```diff
@@ -29,39 +39,132 @@ type CredentialType string
...
+		return ErrMissingField("type")
+	}
 	err := req.Type.Validate()
 	if err != nil {
 		return err
 	}
+	if *req.Type == CredentialTypeAWSSTS && req.Proof == nil {
+		return ErrMissingField("proof")
+	}
+	fingerprint, err := CredentialFingerprint(*req.Type, req.Verifier)
```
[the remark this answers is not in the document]
jpcoenen: The reason for moving away from `CredentialType.Fingerprint(verifier []byte)` is that the current function unifies the fingerprint generation for all credential types (see that function). Though I agree that there is improvement possible.

internals/api/credential.go
```diff
@@ -29,39 +39,132 @@ type CredentialType string
...
+	Request []byte  `json:"request"`
+}
 
+// Validate whether the CredentialProofAWSSTS is valid.
+func (p CredentialProofAWSSTS) Validate() error {
+	if p.Region == nil {
+		return ErrMissingField("region")
+	}
+	if p.Request == nil {
+		return ErrMissingField("request")
```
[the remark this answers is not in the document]
jpcoenen: That's not realistically possible. This is a complete request, including headers.

internals/api/credential.go
```diff
@@ -29,39 +39,132 @@ type CredentialType string
...
+
+// CredentialFingerprint returns the fingerprint of a credential.
+func CredentialFingerprint(t CredentialType, verifier []byte) (string, error) {
+	var toHash []byte
+	if t == CredentialTypeRSA {
+		// Provide compatibility with traditional RSA credentials.
+		toHash = verifier
+	} else {
+		encodedVerifier := base64.RawStdEncoding.EncodeToString(verifier)
+		toHash = []byte(fmt.Sprintf("credential_type=%s;verifier=%s", t, encodedVerifier))
```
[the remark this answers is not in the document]
jpcoenen: Technically, yes, you're right, it should work. But since we're using a string format as a "container", it is nice to keep everything a string. 

Not applicable in this case, but applying the strategy you suggest can in some cases lead to collision attacks. For example, if we had the following structure: `a=<value-of-a>;b=<value-of-b>`, I can think of a few value for `a` and `b` that produce a collision. For example:

a: `1;b=2`
b: `3`

Produces the same result as:
a: `1`
b: `2;b=3`

Both would produce `sha256(`a=1;b=2;b=3`)` as a result. Not good.

In our case `credential_type` is limited to a few values. But still it is a good practice to not open up for these kinds of things. That also why (though not strictly necessary), I used `base64.RawStdEncoding`. This omits the `=` as padding.

internals/api/encryption_key.go
```diff
@@ -0,0 +1,262 @@
...
+func (EncryptionKeyEncrypted) AlgorithmSupported(a EncryptionAlgorithm) bool {
+	return a == EncryptionAlgorithmAESGCM || a == EncryptionAlgorithmRSAOEAP
+}
+
+// Validate whether the EncryptionKeyEncrypted is valid.
+func (k EncryptionKeyEncrypted) Validate() error {
+	if k.Length == nil {
+		return ErrMissingField("length")
+	}
+	if IntValue(k.Length) <= 0 {
```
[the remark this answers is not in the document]
jpcoenen: Doing some more sanity checks on used key lengths might be a good idea, yes. Though this one still is very thin if you did check for any of the proposed values because technically a user of the API can provide any value he likes here.

internals/api/encryption_key.go
```diff
@@ -0,0 +1,262 @@
...
+// EncryptionKeyEncrypted is an encryption key that has been encrypted by another key.
+type EncryptionKeyEncrypted struct {
+	EncryptionKey
+	Length       *int           `json:"length"`
+	EncryptedKey *EncryptedData `json:"encrypted_key"`
+}
+
+// AlgorithmSupported returns whether the given EncryptionAlgorithm is supported by this type of encrytpion key.
+func (EncryptionKeyEncrypted) AlgorithmSupported(a EncryptionAlgorithm) bool {
+	return a == EncryptionAlgorithmAESGCM || a == EncryptionAlgorithmRSAOEAP
```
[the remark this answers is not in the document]
jpcoenen: We could do a `strings.EqualFold()`, yes.

internals/api/encryption_key.go
```diff
@@ -0,0 +1,262 @@
...
+	return &EncryptionKeyLocal{
+		EncryptionKey: EncryptionKey{
+			Type: KeyTypeLocal,
+		},
+		Length: Int(length),
+	}
+}
+
+// EncryptionKeyLocal is an encryption key that has is stored locally by the user.
+type EncryptionKeyLocal struct {
```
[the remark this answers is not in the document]
jpcoenen: Jep.

internals/api/encryption_key.go
```diff
@@ -0,0 +1,262 @@
...
+}
+
+// Validate whether the KeyDerivationMetadataScrypt is valid.
+func (KeyDerivationMetadataScrypt) Validate() error {
+	return nil
+}
+
+// UnmarshalJSON populates an EncryptionKeyDerived from a JSON representation.
+func (ek *EncryptionKeyDerived) UnmarshalJSON(b []byte) error {
+	return errors.New("derived key type not yet supported")
```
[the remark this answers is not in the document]
jpcoenen: We could remove the code for now altogether. The main reason for adding the code was to check whether everything would work in the API.

internals/auth/nop.go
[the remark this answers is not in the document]
jpcoenen: We could the `httpClient` to not need this. However, this was by far the easiest method to get stuff working; and it works like a charm.

Making it unexported sounds like a great plan.

internals/aws/kms_decrypter.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/aws/aws-sdk-go/service/kms/kmsiface"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/secrethub/secrethub-go/internals/api"
+)
+
+// KMSDecrypter is an implementation of the secrethub.Decrypter interface that uses AWS KMS for decryption.
+type KMSDecrypter struct {
+	kmsSvcGetter func(region string) kmsiface.KMSAPI
```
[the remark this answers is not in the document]
jpcoenen: It means service, indeed. It is taken directly from the `aws-sdk-go` samples.

internals/aws/kms_decrypter.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// NewKMSDecrypter returns a new KMSDecrypter that uses the provided configuration to configure the AWS session.
+func NewKMSDecrypter(cfgs ...*aws.Config) (*KMSDecrypter, error) {
```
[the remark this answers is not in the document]
jpcoenen: As far as I know, the same thing as calling it without arguments; which works fine.

internals/aws/kms_decrypter.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/service/kms/kmsiface"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/secrethub/secrethub-go/internals/api"
+)
+
+// KMSDecrypter is an implementation of the secrethub.Decrypter interface that uses AWS KMS for decryption.
+type KMSDecrypter struct {
```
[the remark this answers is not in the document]
jpcoenen: I'm fine with any of the two.

internals/aws/service_creator.go
```diff
@@ -0,0 +1,141 @@
...
+
+import (
+	"bytes"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/service/kms/kmsiface"
+	"github.com/aws/aws-sdk-go/service/sts/stsiface"
+
+	"github.com/aws/aws-sdk-go/service/sts"
+
```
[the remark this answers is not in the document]
jpcoenen: It's one of the few dislikes I have about GoLand; it adds imports in weird places.

internals/aws/service_creator.go
```diff
@@ -0,0 +1,141 @@
...
+	err = encryptReq.HTTPRequest.Write(&buf)
+	if err != nil {
+		return nil, handleError(err)
+	}
+	return buf.Bytes(), nil
+}
+
+// parseRole tries to parse an inputted role into a role ARN.
+// The input can either be an ARN or the name of a role (prefixed with role/ or not)
+// The outputted value is not guaranteed to be a valid ARN.
```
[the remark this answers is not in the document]
jpcoenen: Suggestion sounds great.

pkg/secrethub/auth.go
jpcoenen: If we add any other cloud services, we could add them here.

pkg/secrethub/auth_aws.go
jpcoenen: It's a combination of awkwardness of the design in combination with using it weirdly. 

`.WithRegion(region)` is to signal to the AWS Client that it should sign it's requests for that region. It also changes the endpoint..... BUT specifically for `STS` the client refuses to use any other endpoint than `sts.amazonaws.com`, which is the _global_ STS endpoint (in the US). We like to use the local endpoints because the `eu-west-1` endpoint is closer to our server. Haven't found another way of achieving this than doing this workaround.

pkg/secrethub/auth_aws.go
jpcoenen: Absolutely agree that the contract can be improved. If we want `client := secrethub.NewClient().WithAWS(foo)` then how do we get a client with file-based credentials? `client := secrethub.NewClient().WithFileCredential("path")`?

pkg/secrethub/auth_aws.go
jpcoenen: From a client-perspective that makes sense. But what would that look like on the API? It should be a POST request, but a `POST /v1/session` doesn't look right. So should we stay with `POST /v1/auth`?

pkg/secrethub/client.go
jpcoenen: That makes sense.

pkg/secrethub/auth_aws.go
jpcoenen: It's borrowed from AWS. We can also rename it to `SessionKey` or just `Key` (it's in a session-related struct anyway).

pkg/secrethub/client.go
jpcoenen: As said before, we could rename the client side of this to session, I can agree with that.

pkg/secrethub/credential.go
jpcoenen: Doesn't that lead to a circular dependency?

pkg/secrethub/credential.go
jpcoenen: You're :100:% right.

pkg/secrethub/http.go
jpcoenen: I think we notice it soon enough if this leads to problems.

pkg/secrethub/http.go
jpcoenen: `http.StatusCreated` sounds good.

pkg/secrethub/service.go
jpcoenen: In hindsight, splitting `verifier` and `encrypter` does indeed seem to be unnecessary.

jpcoenen: > Also, how do you feel about having the interface composed this way with a separate AWS interface but having the Create function on the 'root' ServiceService interface?

What do you mean?

internals/api/auth.go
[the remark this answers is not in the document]
jpcoenen: Rephrased to `could not find an AWS endpoint for the provided region`

internals/api/auth.go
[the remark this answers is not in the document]
jpcoenen: Rephrased to `encountered an unexpected problem while verifying your identity on AWS. Please try again later.`. What do you think?

internals/api/auth.go
```diff
@@ -0,0 +1,249 @@
...
+	ErrInvalidPayload      = errAPI.Code("invalid_payload").StatusError("invalid payload provided for authentication request", http.StatusBadRequest)
+	ErrInvalidAuthMethod   = errAPI.Code("invalid_auth_method").StatusError("invalid auth method", http.StatusBadRequest)
+	ErrMissingField        = errAPI.Code("missing_field").StatusErrorPref("request is missing field %s", http.StatusBadRequest)
+	ErrSessionNotFound     = errAPI.Code("session_not_found").StatusError("session could not be found, it might have expired", http.StatusForbidden)
+	ErrSessionExpired      = errAPI.Code("session_expired").StatusError("session has expired", http.StatusForbidden)
+	ErrAuthFailed          = errAPI.Code("auth_failed").StatusError("authentication failed", http.StatusForbidden)
+	ErrCouldNotGetEndpoint = errAPI.Code("wrong_endpoint").StatusError("could not find the AWS endpoint", http.StatusBadRequest)
+	ErrAWSException        = errAPI.Code("aws_exception").StatusError("AWS returned an error", http.StatusFailedDependency)
+	ErrNoServiceWithRole   = errAPI.Code("no_service_with_role").StatusErrorPref("there exists no service account tied to the AWS IAM role '%s'", http.StatusNotFound)
```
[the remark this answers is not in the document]
jpcoenen: :heavy_check_mark:

internals/api/auth.go
```diff
@@ -0,0 +1,249 @@
...
+		Payload: &SessionPayloadHMAC{
+			SecretKey: &secretKey,
+		},
+	}
+}
+
+// Session represents a session that can be used for authentication to the server.
+type Session struct {
+	SessionID  *uuid.UUID   `json:"session_id"`
+	Expiration *time.Time   `json:"expiration"`
```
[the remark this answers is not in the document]
jpcoenen: Are you okay with `Expires`?  `ExpiresAt` feels like it refers to a location.

internals/api/credential.go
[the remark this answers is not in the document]
jpcoenen: Doesn't ` Make sure you are using AWS credentials that correspond to the role you are trying to add.` clarify it enough?

internals/api/encryption_key.go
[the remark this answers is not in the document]
jpcoenen: I am now converting all these types to lowercase on Unmarshal.

pkg/secrethub/fakeclient/client.go
```diff
@@ -24,6 +25,11 @@ func (c Client) Accounts() secrethub.AccountService {
 	return c.AccountService
 }
 
+// Auth implements the secrethub.Client interface.
+func (c Client) sessions() secrethub.SessionService {
```
golangcibot: U1000: func `Client.sessions` is unused (from `unused`)

golangcibot: U1000: func `Client.sessions` is unused (from `unused`)

golangcibot: U1000: func `Client.sessions` is unused (from `unused`)

pkg/secrethub/client.go
```diff
@@ -13,7 +13,7 @@ import (
 type Client interface {
 	AccessRules() AccessRuleService
 	Accounts() AccountService
-	Auth() AuthService
```
mackenbach: If it's private, should it be part of the interface or can it just be a function on the concrete implementation of the interface?

mackenbach: Also the return type is public, should that be private too?

internals/api/auth.go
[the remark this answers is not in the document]
mackenbach: I know, but let's look at this from the user's perspective. The most likely case when they see this error is that their session expired. Wording it like the possible expiration is a consequence of the not found error unnecessarily exposes an implementation detail to the user that they don't need to be aware of. Being vague in our language would only confuse them more rather than inform them. 

`session not found or expired` makes kinda sense IMO, another alternative is `no valid session found`. 

Anyway, it's not a super big thing. Proceed as you see fit.

internals/api/auth.go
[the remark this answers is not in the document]
mackenbach: Nice, that's better. Consider calling the error code `wrong_aws_endpoint` as it is specific to AWS.

internals/api/auth.go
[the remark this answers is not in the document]
mackenbach: Hmm interesting. So could this happen when an account doesn't have permissions to call AWS STS? If the case never happens, it's redundant to have a specific error for it. If it does, it's good to keep it and make sure we have the correct error code.

internals/api/auth.go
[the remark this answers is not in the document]
mackenbach: `Expires` is cool with me. Feels a bit like a `bool` though: `if session.Expires { ... }`. We use quite a lot of `xxx_at` patterns already (e.g. `created_at`, `last_modified_at`). 

I don't have the association you have with `at` referring to a location. Anyways, it's okay with me whichever you choose.

internals/api/auth.go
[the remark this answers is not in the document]
mackenbach: Yes checking whether the value is actually of the right format would be better than only checking if it's not-nil.

internals/api/credential.go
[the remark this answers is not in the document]
mackenbach: Then consider this error message: `the AWS Account ID in the role ARN does not match the AWS Account ID of the AWS credentials used for authentication. Make sure you are using AWS credentials that correspond to the role you are trying to add.`

internals/api/credential.go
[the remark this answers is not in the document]
mackenbach: Alright so we keep it a separate function not attached to a certain type. Then how do you feel about naming it `GetFingerprint(*req.Type, req.Verifier)`?

internals/aws/kms_decrypter.go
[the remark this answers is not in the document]
mackenbach: Hmmm ok. If it's consistent with how AWS uses it I guess there are benefits to be consistent as well.

internals/aws/kms_decrypter.go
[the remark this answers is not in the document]
mackenbach: Hmm I don't know how the aws SDK handles this, but by default variadic pointer arguments can cause some weird stuff: https://play.golang.org/p/4q5JaVf8264

pkg/secrethub/auth_aws.go
mackenbach: Oh wow, that's interesting. There's actually a very good reason for doing it this awkwardly 🎉  Consider adding a small comment to explain what were doing here so we'll still know the rationale when someone comes along in 6 months and thinks "hey this is weird, it can be simpler".

pkg/secrethub/auth_aws.go
mackenbach: Jotted down some notes as to how we could use functional options like we discussed in person. Not really sure whether I like the `client.With(opts)` function but it has a certain elegance. It's pretty similar to the pattern a lot of Golang router packages use, e.g. `chi.Router.Use`.


```go
package secrethub

type Option func(c Client) (Client, error)

func NewClient(opts ...Option) (Client, error) {
   ...
   c := foo 
   for _, opt := range opts {
      c, err = opt(c)
      if err != nil {
         return nil, err
      }
   }

   return c, nil
}

func (c client) With(opts ...Option) (Client, error) { ... }

func WithServerURL(url url.URL) Option {...}

func WithTransport(t http.Transport) Option {...}

func WithAWS(foo string) Option {
   return func(c Client) (Client, error) {
      // do aws stuff
      ...
   }
} 

func WithCredential(raw string, passphrase string) Option {
   return func(c Client) (Client, error) {
      // parse credential etc
      ...
   }
}

func WithCredentialFile(path string) Option {}
```

So you could call it something like:

```go
// Create a new client
client, err := secrethub.NewClient(
   secrethub.WithAWS(foo), 
   secrethub.WithServerURL("https://api.secrethub.io"),
)

// Change it on the fly, e.g. for a handful of requests
specialClient, err = client.With(secrethub.WithTransport(t))
```

pkg/secrethub/auth_aws.go
mackenbach: `POST /v1/sessions` makes quite some sense to me. Then you could also (later) build `GET /v1/sessions` to see all your currently logged in devices or something similar. 

Then `CreateSession` makes a lot of sense for the function name. If we ever want to introduce the concept of sessions as an API resource (e.g. similar to a repository being a resource) instead of an authentication thingy, we could now decide to put the `/sessions` route under e.g. `/v1/auth/sessions` or `/v1/me/sessions` so it won't conflict later.

I don't foresee this being a problem anytime soon though.

pkg/secrethub/auth_aws.go
mackenbach: I'd say stick to `SessionKey` or `Key` whichever you prefer.

pkg/secrethub/credential.go
mackenbach: Hmm ok I agree with your rule of thumb to keep the two separate. Don't like the hard coded 8 here though.

pkg/secrethub/service.go
mackenbach: Wow hard to read question I formulated there. Wha tI meant is we have a 'plain` service that will be created on `client.Services().Create()` and an AWS service that will be created on a level of nesting deeper: `client.Services().AWS().Create()`. 

My question is, should we then make 'plain' services be created on e.g. (and this is far from pretty) `client.Services().Default().Create()`?

pkg/secrethub/fakeclient/client.go
golangcibot: U1000: func `Client.sessions` is unused (from `unused`)

golangcibot: U1000: func `Client.sessions` is unused (from `unused`)

golangcibot: U1000: func `Client.sessions` is unused (from `unused`)

golangcibot: U1000: func `Client.sessions` is unused (from `unused`)

internals/api/auth.go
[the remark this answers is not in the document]
jpcoenen: _Everyone_ has permissions to call `GetCallerIdentity`: 

>No permissions are required to perform this operation. If an administrator adds a policy to your IAM user or role that explicitly denies access to the sts:GetCallerIdentity action, you can still perform this operation. Permissions are not required because the same information is returned when an IAM user or role is denied access. To view an example response, see I Am Not Authorized to Perform: iam:DeleteVirtualMFADevice. 

If you are disallowed to do something, an `AwsAccessDeniedException` is returned.

pkg/secrethub/fakeclient/client.go
golangcibot: U1000: func `Client.sessions` is unused (from `unused`)

pkg/secrethub/auth_aws.go
jpcoenen: `POST /v1/sessions` is not correct from a REST point of view. It means: "I have a session here, save it for me.". But that's not what we're doing. You're sending authentication data and retrieving a newly created session.

What about the following?
`POST /v1/auth` -> authenticate and create a session
`GET /v1/auth/sessions` -> returns all user's sessions.
`DELETE /v1/auth/sessions/<id>` -> retract a session

pkg/secrethub/auth_aws.go
jpcoenen: That conflicts with an import ;-)

SimonBarendse:
Only thing left is implementing proof for RSA credentials, so that we can make unauthenticated user request (we currently sign the request with the credential to provide proof of possession).

internals/api/auth.go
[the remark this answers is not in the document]
SimonBarendse: We've decided to go with expires_at

internals/aws/kms_decrypter.go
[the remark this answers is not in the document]
jpcoenen: As this AWS accepts them, I think we should not worry about this. We only pass these to the AWS sdk.

internals/aws/kms_decrypter.go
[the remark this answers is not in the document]
jpcoenen: I choose to leave it as is.

internals/aws/service_creator.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/kms"
+	"github.com/secrethub/secrethub-go/internals/api"
+)
+
+// ServiceCreator is an implementation of the secrethub.Verifier and secrethub.Encrypter interface that can be used
+// to create an AWS service account.
+type ServiceCreator struct {
+	stsSvc        stsiface.STSAPI
+	kmsSvc        kmsiface.KMSAPI
```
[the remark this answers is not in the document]
jpcoenen: Previously discussed.

pkg/secrethub/account_key.go
```diff
@@ -15,7 +15,7 @@ func generateAccountKey() (crypto.RSAPrivateKey, error) {
 // AccountKeyService handles operations on SecretHub account keys.
 type AccountKeyService interface {
 	// Create creates an account key for the client's credential.
-	Create(encrypter Encrypter) (*api.EncryptedAccountKey, error)
+	Create(credentialFingerprint string, encrypter Encrypter) (*api.EncryptedAccountKey, error)
```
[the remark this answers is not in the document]
jpcoenen: @SimonBarendse for you

pkg/secrethub/auth.go
jpcoenen: Removed

pkg/secrethub/auth_aws.go
jpcoenen: @SimonBarendse for you

pkg/secrethub/auth_aws.go
jpcoenen: Discussed this: consistent with AWS.

pkg/secrethub/auth_aws.go
jpcoenen: @mackenbach, can I resolve this?

pkg/secrethub/auth_aws.go
jpcoenen: @SimonBarendse still open

pkg/secrethub/auth_aws.go
jpcoenen: @SimonBarendse still open

pkg/secrethub/auth_aws.go
jpcoenen: No clue what this means.

pkg/secrethub/account_key.go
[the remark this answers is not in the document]
jpcoenen: The cut currently is at: everything that is an English word is spelled using correct English spelling. Otherwise `+er` is used.

pkg/secrethub/auth_aws.go
mackenbach: Yes `/v1/auth/sessions` as an endpoint seems fine to me. Whether you then use `/v1/auth` or `/v1/auth/sessions` for the POST request is a bit preference. One thing to think about is that we may want to add `/v1/auth/foo_method` or something later, so consider how that impacts adding the POST on the `/v1/auth` endpoint.

mackenbach: My comment was mostly about the client name, which I think was renamed to `CreateSession` or something like that. So if that's correct, you can resolve this discussion as far as I'm concerned.

pkg/secrethub/auth_aws.go
mackenbach: Me neither. It feels like I was continuing a conversation I had with myself in another comment, but I cannot find anything that corresponds to this. One thing that catches my eye that I could have meant is: are all `AuthRequestsAWSSTS` of type `HMAC`? In other words, do we want to expose the option to set the `SessionType` with the function or does it default to something?

Tiny thing. Don't spend too much time on it. Resolve it whenever you're satisfied.

jpcoenen: If we do not want to break this in the future, we have to give it as an option. There currently is only one option.. so it should not be difficult to choose one. Furthermore, I don't think many users will interact with this.

So I am choosing to leave it as is.

PR #94, opened 2019-08-05
Remove billing functionality

jpcoenen:
This is not exposed for development. Removing it from the public API
package makes this clear. It also forces a clear separation between
billing and other business logic.


SimonBarendse:
>One question before merging: what happens when the server still returns the subscription field and the client is on the new version which doesn't know this field exists?

@mackenbach It will be ignored.
See: https://golang.org/pkg/encoding/json/#Unmarshal

> By default, object keys which don't have a corresponding struct field are ignored

This is also why we can add fields to the JSON response without breaking old versions of the client (that don't know about the existence of this field).

mackenbach:
Thanks @SimonBarendse for the answer. Follow-up question would be: how does the javascript client handle this? 

Also, what happens when the converse case occurs? Clients expect a `subscription` field, but it's not there?

jpcoenen:
> Also, what happens when the converse case occurs? Clients expect a subscription field, but it's not there?
Then it's set to the empty value, but the field is used nowhere by us in Go. And I think it is safe to assume no user of the client is using it. Do you agree?

>  Follow-up question would be: how does the javascript client handle this?
Let's discuss this somewhere else.

SimonBarendse:
@mackenbach Good questions.

I'm not sure what the JavaScript client does. Let me check and get back to you.

As for the converse case, a field that is expected but not supplied in the JSON: It will not be set by the JSON library. So it will be the zero-value (`nil` in the case of `subscription`).
Note that this is a breaking change on the server, as we remove existing functionality.

mackenbach:
Alright, a zero value that we don't use in Go code is acceptable I'd say for the converse (Go) case. Discussed in person the Javascript side. We're going to make sure the release roll-out goes smoothly together across languages.

PR #95, opened 2019-08-08
Client initialisation improvement

jpcoenen:
This is a demonstration of what initialising the client could look like. For example:

```go
client := secrethub.Must(secrethub.NewClient())

secret, err := client.Secrets().Get("path/to/secret")
[..]
```

```go
client := secrethub.Must(secrethub.NewClient(secrethub.Credentials(secrethub.AWS()), secrethub.Remote("https://my-secrethub-remote")))

secret, err := client.Secrets().Get("path/to/secret")
[..]
```

I am not satisfied with how long it is, mainly because everything is in the `secrethub` package now. Things could be nicer if we moved some things, e.g.:

```go
client := secrethub.Must(secrethub.NewClient(opt.Credentials(cred.AWS()), opt.Remote("https://my-secrethub-remote")))

secret, err := client.Secrets().Get("path/to/secret")
[..]
```

However, this is currently not possible because of the import cycles this would introduce. So if this is what we want, we should reorganise some code to make this possible.

This is mainly meant as an opportunity for a discussion. So let me know your thoughts.


pkg/secrethub/credential_providers.go
```diff
@@ -0,0 +1,27 @@
...
+		}
+		authenticator, err := c.Sessions().AWS(awsCfg...).Create()
+		if err != nil {
+			return nil, nil, err
+		}
+		return authenticator, decrypter, nil
+	}
+}
+
+func RSA(credential *RSACredential) CredentialProvider {
```
golangcibot: `credential` can be `github.com/secrethub/secrethub-go/internals/auth.Signer` (from `interfacer`)

mackenbach:
I don't really like the `secrethub.Must` wrapper. It's fairly ugly and encourages bad initialization. It's nice for example code to remove the error check, but for quick examples we can also just write: 

```go
client, err := secrethub.NewClient(...)
secret, err := client.Secrets().Get("path/to/secret")
```

Which implies there is an error that we're not checking. 

Also, I don't really like the `secrethub.Credentials(...)` option. We're exposing the fact that we can reuse this functionality for different credentials by having the user wrap it. I propose to invert this pattern and internally reuse the code that sets the credential (if we want to) e.g.:

```go
func AWS(awsCfg ...*awssdk.Config) ClientOption {
	return func(c *client) error {
		decrypter, err := aws.NewKMSDecrypter(awsCfg...)
		if err != nil {
			return err
		}
		authenticator, err := c.Sessions().AWS(awsCfg...).Create()
		if err != nil {
			return err
		}
	        client.setCredential(authenticator, decrypter)
                return nil
	}
}

func (c *client) setCredential(authenticator auth.Authenticator, decryptor Decrypter) {
		c.decrypter = decrypter
		c.httpClient.authenticator = authenticator
}
```

With these simplifications, I don't feel the names get too long, especially if you space them on multiple lines:

```go
client, err := secrethub.NewClient(
      secrethub.AWS(), 
      secrethub.Remote("https://my-secrethub-remote"),
)

secret, err := client.Secrets().Get("path/to/secret")
[..]
```

What do you think?

mackenbach:
Also, it's important to make clear whether calling an option on the client modifies the original client or returns a new copy of the client. If it can only be used on init (e.g. in the `NewClient` call), then modifying the original client is fine, but if a user of the library can actually call `client.WithOption(myoption)` then that would modify the original client and potentially cause unexpected side effects. 

What do you feel is the best way to go? (no pun intended)

SimonBarendse:
>Also, it's important to make clear whether calling an option on the client modifies the original client or returns a new copy of the client. If it can only be used on init (e.g. in the NewClient call), then modifying the original client is fine, but if a user of the library can actually call client.WithOption(myoption) then that would modify the original client and potentially cause unexpected side effects.
>
>What do you feel is the best way to go? (no pun intended)

@mackenbach

I'd prefer the options to be supplied on construction only. Re-configuring later can have unintended side-effects, such as cached results from one remote being returned after re-configuring the remote.
When we find strong use-cases for modifying existing clients we can always add functionality to do so later and then we will deal with problems this introduces, such as clearing the cache.

Note that the current proposal already allows the re-use of options (e.g. by creating a function that returns these or by storing them in a variable), so that could help when the goal is to make requests with similarly but not equally configured clients.

pkg/secrethub/http.go
```diff
@@ -97,29 +97,18 @@ type httpClient struct {
 
 // newHTTPClient configures a new httpClient and overrides default values
 // when opts is not nil.
-func newHTTPClient(authenticator auth.Authenticator, opts *ClientOptions) *httpClient {
+func newHTTPClient() *httpClient {
```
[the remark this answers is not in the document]
mackenbach: Oh wow that's actually pretty nasty yes. I feel because we split them up into two separate arguments, it should support it if one or the other is not configured/provided. 

Also there are a handful of API calls you can make without authentication, so I feel you should be able to create a client without authentication.

mackenbach:
Also, I like `secrethub.UseAWS()` better than `secrethub.AWS()`.

pkg/secrethub/http.go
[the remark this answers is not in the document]
SimonBarendse: Oke, so as you say above and as we have discussed in person at the office today, we'd like the authenticater and decrypter to be optional on the client constructor.

As the client has functions that don't require a decrypter and the http client has functions that don't require authentication, these can be omitted there as well (so are optional to configure).

For the functions that need them they are required. So an error should be returned when they are not present (in a first implementation this can depend on the server returning an unauthenticated response).

pkg/secrethub/client.go
```diff
@@ -23,127 +21,100 @@ type Client interface {
...
+	errClient = errio.Namespace("client")
+)
 
-// Encrypter encrypts data, typically an account key.
-type Encrypter interface {
-	// Wrap encrypts data, typically an account key.
-	Wrap(plaintext []byte) (*api.EncryptedData, error)
-}
+// Client is a client for the SecretHub HTTP API.
+type Client struct {
```
SimonBarendse: If we make this the client type, I think we should make `ListDirAccounts` private.

pkg/secrethub/client.go
```diff
@@ -1,19 +1,17 @@
...
-	"github.com/secrethub/secrethub-go/internals/aws"
 	"github.com/secrethub/secrethub-go/internals/crypto"
 	"github.com/secrethub/secrethub-go/internals/errio"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/credentials"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/http"
 )
 
 // Client is the SecretHub client.
-type Client interface {
+type ClientAdapter interface {
```
SimonBarendse: As packages should consume the interface and not the struct, I'd prefer to use the nice name `secrethub.Client` for the interface.

Also, the current changeset can have some none intuitive error messages, as `secrethub.Client` now points to the struct (without pointer) instead of the interface.

pkg/secrethub/http.go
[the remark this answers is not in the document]
SimonBarendse: #100

pkg/secrethub/http/options.go
```diff
@@ -0,0 +1,32 @@
+package http
+
+import (
+	"net/http"
+	"time"
+)
+
+type Option func(*Client)
```
mackenbach: Sweet! 😍 Needs comments though.

mackenbach: Also, any specific reason for calling this `Option` as opposed to `ClientOption` which we also use in the `secrethub` package?

pkg/secrethub/main_test.go
```diff
@@ -40,19 +42,20 @@ func init() {
...
-	handler.Handle(baseURLPath+"/", http.StripPrefix(baseURLPath, router))
+	handler.Handle("/v1/", http.StripPrefix("/v1", router))
 
 	// server is a test HTTP server used to provide mock API responses.
 	server := httptest.NewServer(handler)
 
-	opts := &ClientOptions{
-		ServerURL: server.URL,
+	opts := []ClientOption{
+		//	WithRemote(server.URL),
```
mackenbach: Why is this commented out?

pkg/secrethub/org_member_test.go
```diff
@@ -83,7 +82,7 @@ func TestGetOrgMember(t *testing.T) {
 			router, opts, cleanup := setup()
 			defer cleanup()
 
-			client := NewClient(cred1, auth.NewHTTPSigner(cred1), opts)
+			client := Must(NewClient(opts...))
```
mackenbach: Sweet, loving this already

pkg/secrethub/service.go
```diff
@@ -3,12 +3,13 @@ package secrethub
...
 	"github.com/secrethub/secrethub-go/internals/api"
 	"github.com/secrethub/secrethub-go/internals/errio"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/credentials"
 )
 
 // ServiceService handles operations on service accounts from SecretHub.
 type ServiceService interface {
 	// Create creates a new service account for the given repo.
-	Create(path string, description string, verifier Verifier, encrypter Encrypter) (*api.Service, error)
+	Create(path string, description string, verifier credentials.Verifier, encrypter credentials.Encrypter) (*api.Service, error)
```
mackenbach: I think we discussed this, but this signature is completely unclear to the user. They won't know where to start when they want to call this function.

pkg/secrethub/user.go
```diff
@@ -4,26 +4,27 @@ import (
...
+	"github.com/secrethub/secrethub-go/pkg/secrethub/credentials"
 )
 
 // UserService handles operations on users from SecretHub.
 type UserService interface {
 	// Me gets the account's user if it exists.
 	Me() (*api.User, error)
 	// Create creates a new user at SecretHub.
-	Create(username, email, fullName string, verifier Verifier, encrypter Encrypter) (*api.User, error)
+	Create(username, email, fullName string, verifier credentials.Verifier, encrypter credentials.Encrypter) (*api.User, error)
```
mackenbach: Same comment as with service creation. Unclear arguments.

pkg/secrethub/http/http.go
```diff
@@ -1,4 +1,4 @@
-package secrethub
+package http
```
mackenbach: Consider moving this package to `internals/http` as it is far from ready to be published and we don't want to spend too much time polishing it now.

pkg/secrethub/http/encoding.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // Errors
 var (
-	ErrWrongContentType = errClient.Code("wrong_content_type").Error("server returned wrong content type in header")
+	ErrWrongContentType = errHTTP.Code("wrong_content_type").Error("server returned wrong content type in header")
```
mackenbach: Question, why don't we call this `errHTTP` variable simply `errors`? Not for now, I know, but it just popped into my head.

pkg/secrethub/credentials/providers.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/secrethub/secrethub-go/internals/api"
+	"github.com/secrethub/secrethub-go/internals/auth"
+	"github.com/secrethub/secrethub-go/internals/aws"
+	"github.com/secrethub/secrethub-go/internals/crypto"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/credentials/sessions"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/http"
+)
+
+// Verifier exports verification bytes that can be used to verify signed data is processed by the owner of a signer.
+type Verifier interface {
```
mackenbach: Are these interfaces in the right file? They seem a bit lost here.

pkg/secrethub/http/http.go
```diff
@@ -87,43 +89,44 @@ type ClientOptions struct {
...
-		version:       ClientVersion,
+		base: serverURL,
+		//version: secrethub.ClientVersion,
 	}
+	client.Options(with...)
+	return client
 }
 
-func (c *httpClient) Authenticate(in interface{}) (*api.Session, error) {
+func (c *Client) Options(with ...Option) {
```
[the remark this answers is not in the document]
SimonBarendse: See also this comment (https://github.com/secrethub/secrethub-go/pull/95#issuecomment-520384922) about what needs to be done if we want to support changing options on existing clients.

pkg/secrethub/http/http.go
SimonBarendse: Can also be `pkg/secrethub/internals/http` if we want it to be a subpackage of `pkg/secrethub`. 

Then it's also actually enforced that this can only be consumed by the secrethub package.

pkg/secrethub/client.go
```diff
@@ -23,127 +21,110 @@ type Client interface {
...
+			return nil, err
+		}
 	}
-	client := &clientAdapter{
-		client: newClient(decrypter, auth.NopAuthenticator{}, opts),
+
+	// Try to use default key credentials if none provided explicitly
+	if client.decrypter == nil {
+		err := WithCredentials(credentials.UseKey(nil, nil))(client)
+		if err != nil {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

pkg/secrethub/credentials/readers.go
```diff
@@ -0,0 +1,54 @@
...
+
+// Errors
+var (
+	ErrCannotFindHomeDir = errCredentials.Code("cannot_find_home_dir").ErrorPref(
+		"cannot find your home directory: %s",
+	)
+)
+
+type Reader interface {
+	Read() ([]byte, error)
```
mackenbach: How is this different from the `io.Reader` interface? I get that `io.Reader` accepts a byte slice as an argument, but that's just implementation detail right?

Please add a clarifying comment as to why we have this interface here.

pkg/secrethub/credentials/readers.go
```diff
@@ -52,3 +55,23 @@ func FromString(raw string) Reader {
 		return []byte(raw), nil
 	})
 }
+
+func fromDefault() Reader {
```
mackenbach: `fromDefault` never returns a passphrase but always a credential wrapped in a `Reader`. So should we then change the name to reflect that meaning?

pkg/secrethub/http/options.go
```diff
@@ -7,7 +7,7 @@ import (
 
 type Option func(*Client)
 
-func WithRemote(remote string) Option {
+func WithServerURL(remote string) Option {
```
mackenbach: Argument could be named `url` too, do we want to validate it, and/or do we want it to be a `url.URL` type?

pkg/secrethub/client.go
```diff
@@ -57,6 +57,16 @@ func NewClient(with ...ClientOption) (*Client, error) {
 			return nil, err
 		}
 	}
+
+	// Try to use default key credentials if none provided explicitly
+	if client.decrypter == nil {
```
mackenbach: Only when the decrypter is `nil`? Or also when the `authenticator` is nil?

pkg/secrethub/credentials/providers.go
```diff
@@ -60,7 +52,7 @@ func UseAWS(awsCfg ...*awssdk.Config) Provider {
 //		credentials.UseKey(credentials.FromBytes("<a credential>"))
 //		credentials.UseKey(credentials.FromFile("~/.secrethub/credential"), credentials.FromString("passphrase"))
 func UseKey(credentialReader Reader, passReader Reader) Provider {
-	return func() (http.AuthProvider, Decrypter, error) {
+	return func(_ *http.Client) (auth.Authenticator, Decrypter, error) {
```
mackenbach: Did we just move the problem one layer up but it's still there?

jpcoenen: I challenge you to solve it differently.

pkg/secrethub/http/options.go
jpcoenen: Missed that rename, fixed.

Using an `url.URL` makes it a lot less friendly to use this function. If someone uses a malformed URL, he'll find out at runtime. Sounds okay to me.

pkg/secrethub/credentials/readers.go
jpcoenen: Agreed with a rename to `credentialFromDefault()`?

pkg/secrethub/credentials/readers.go
jpcoenen: I'll have a look again. The implementations might get a little complicated though.

jpcoenen: `readerFunc()` indeed becomes a little bit more complicated. But it works :-)

pkg/secrethub/client.go
jpcoenen: Theoretically, yes. But... that is not accessible from the `secrethub` package. Since they should always be set at the same time, in practice the same effect is achieved. Not perfectly nice, but I'd say let's go with it for now and fix it if any problems arise from it.

pkg/secrethub/credentials/encoding_test.go
```diff
@@ -223,16 +223,16 @@ func TestParser(t *testing.T) {
...
-	cred, err := generateRSACredential(1024)
+	cred, err := GenerateRSACredential(1024)
 	assert.OK(t, err)
 
-	parser := NewCredentialParser(DefaultCredentialDecoders)
+	parser := newCredentialParser(DefaultCredentialDecoders)
 
 	// Act
-	raw, err := EncodeCredential(cred)
+	raw, err := encodeCredential(cred)
```
golangcibot: undeclared name: `encodeCredential` (from `typecheck`)

pkg/secrethub/http/encoding.go
jpcoenen: `errors` collides with the package named `errors`.

pkg/secrethub/http/http.go
```diff
@@ -636,9 +639,16 @@ func (c *httpClient) do(rawURL string, method string, expectedStatus int, in int
...
-	err = c.authenticator.Authenticate(req)
-	if err != nil {
-		return errio.Error(err)
+	if c.authProvider != nil {
+		authenticator, err := c.authProvider.Provide(c)
+		if err != nil {
+			return errio.Error(err)
+		}
+
+		err = authenticator.Authenticate(req)
```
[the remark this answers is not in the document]
jpcoenen: Removed it.

pkg/secrethub/credentials/sessions/auth_provider.go
```diff
@@ -0,0 +1,28 @@
+package sessions
+
+import (
+	"github.com/secrethub/secrethub-go/internals/auth"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/http"
+)
+
+func NewAuthProvider(sessionCreator SessionCreator) *AuthProvider {
```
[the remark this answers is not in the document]
jpcoenen: Removed it.

pkg/secrethub/credentials/providers.go
```diff
@@ -0,0 +1,102 @@
...
+type Encrypter interface {
+	// Wrap encrypts data, typically an account key.
+	Wrap(plaintext []byte) (*api.EncryptedData, error)
+}
+
+type staticCredentialAuthProvider struct {
+	credential auth.Signer
+}
+
+func (s staticCredentialAuthProvider) Provide(_ *http.Client) (auth.Authenticator, error) {
```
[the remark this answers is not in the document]
jpcoenen: Removed it.

pkg/secrethub/credentials/encoding.go
```diff
@@ -0,0 +1,277 @@
...
+
+// Name returns the encoding name.
+func (d RSAPrivateKeyDecoder) Name() string {
+	return "rsa"
+}
+
+// CredentialDecoder converts a payload into a Credential.
+type CredentialDecoder interface {
+	// Decode decodes a payload into a Credential.
+	Decode(payload []byte) (*RSACredential, error)
```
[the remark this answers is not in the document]
jpcoenen: My idea was that this does not have any impact on the user contract. So I propose to leave it as is.

pkg/secrethub/credentials/sessions/session_aws.go
```diff
@@ -39,16 +37,20 @@ func (s awsSessionService) Create() (auth.Authenticator, error) {
...
-	resp, err := s.client.httpClient.Authenticate(req)
+	resp, err := httpClient.Authenticate(req)
 	if err != nil {
 		return nil, err
 	}
 	if resp.Type != api.SessionTypeHMAC {
 		return nil, api.ErrInvalidSessionType
 	}
-	hmacSession := resp.HMAC()
+	sess := resp.HMAC()
```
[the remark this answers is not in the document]
jpcoenen: My IDE starts whining when you do this because of the colission with the aws `session` package.

pkg/secrethub/main_test.go
jpcoenen: Because it did not work yet :upside_down_face:

pkg/secrethub/http/options.go
mackenbach: Alright, other issue then is how do we return an error when the server cannot be reached? I think that error message looks really weird.

pkg/secrethub/credentials/creators.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	c.key = key
+	return c.key, c.key, nil
+}
+
+func (c *KeyCreator) Export() []byte {
+	if c.key == nil {
+		return nil
```
mackenbach: Should this return `nil` or a zero `[]byte{}`?

pkg/secrethub/credentials/creators.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	return c.key.Export()
+}
+
+func CreateAWS(kmsKeyID string, roleARN string, awsCfg ...*awssdk.Config) Creator {
+	return creatorFunc(func() (Verifier, Encrypter, error) {
+		creator, err := aws.NewServiceCreator(kmsKeyID, roleARN, awsCfg...)
```
mackenbach: Is the name `aws.NewServiceCreator` still correct, or should it be `aws.NewCredentialCreator`?

pkg/secrethub/credentials/encoding.go
```diff
@@ -43,7 +43,7 @@ type EncodableCredential interface {
 	// Export exports the credential in a format that can be decoded by its Decoder.
 	Export() []byte
 	// Decoder returns a decoder that can decode an exported key back into a Credential.
-	Decoder() CredentialDecoder
+	Decoder() credentialDecoder
```
mackenbach: Not sure, but I think the linter won't like this. Returning a private type from a public function is generally not accepted.

pkg/secrethub/credentials/encoding.go
```diff
@@ -28,12 +28,12 @@ var (
...
-	// DefaultCredentialDecoders defines the default list of supported decoders.
-	DefaultCredentialDecoders = []credentialDecoder{rsaPrivateKeyDecoder{}}
-	// DefaultCredentialEncoding defines the default encoding used for encoding credential segments.
-	DefaultCredentialEncoding = base64.URLEncoding.WithPadding(base64.NoPadding)
+	// DefaultParser defines the default parser for credentials.
+	DefaultParser = newParser(DefaultDecoders)
+	// DefaultDecoders defines the default list of supported decoders.
+	DefaultDecoders = []decoder{rsaPrivateKeyDecoder{}}
+	// DefaultEncoding defines the default encoding used for encoding credential segments.
+	DefaultEncoding = base64.URLEncoding.WithPadding(base64.NoPadding)
```
mackenbach: I think some of these variables could very well be private?

pkg/secrethub/client.go
jpcoenen: Then it should always have been private. But sure. No reason to export it.

pkg/secrethub/client.go
jpcoenen: Can you please clarify your second point? (example can suffice)

pkg/secrethub/http/options.go
jpcoenen: That's a good point. I will create a separate PR for that when this is done. Will also include client timeouts in that.

pkg/secrethub/credentials/creators.go
jpcoenen: For almost all intents and purposes that is the same. Only difference I know is how they're JSON-encoded. In that case `nil` makes more because it is "not set", the key is not of length 0.

Altenative would be to add an error to this function.

pkg/secrethub/credentials/creators.go
jpcoenen: :heavy_check_mark:

pkg/secrethub/credentials/sessions/session.go
```diff
@@ -0,0 +1,37 @@
...
+type Session interface {
+	NeedsRefresh() bool
+	Authenticator() auth.Authenticator
+}
+
+type hmacSession struct {
+	sessionID  uuid.UUID
+	sessionKey string
+
+	expireTime
```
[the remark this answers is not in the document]
jpcoenen: Because it has an expire time and by embedding the type, all functions. I actually found this a nice approach.

pkg/secrethub/http/http.go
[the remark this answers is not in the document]
jpcoenen: The reason is that it makes the usage a lot more straightforward. Otherwise we first have to gather all options that concern the HTTP Client before constructing the HTTP Client. With this function we can just create an HTTP Client and then add the options incrementally. In practice this is only done on init (because only the `secrethub` package has access to the HTTP Client and only calls `Options()` on init.) So this should not cause any problems. We could try to rewrite it, but I think it would become overly complex for something that is not exposed to the user (`http` is in `internals/`).

`With` and `WithOptions` indicate that it _returns_ a new client (at least to me). `ApplyOptions` okay?

pkg/secrethub/credentials/encoding.go
jpcoenen: Once again, very true.

pkg/secrethub/credentials/encoding.go
```diff
@@ -0,0 +1,277 @@
...
+var (
+	// DefaultCredentialParser defines the default parser for credentials.
+	DefaultCredentialParser = NewCredentialParser(DefaultCredentialDecoders)
+	// DefaultCredentialDecoders defines the default list of supported decoders.
+	DefaultCredentialDecoders = []CredentialDecoder{RSAPrivateKeyDecoder{}}
+	// DefaultCredentialEncoding defines the default encoding used for encoding credential segments.
+	DefaultCredentialEncoding = base64.URLEncoding.WithPadding(base64.NoPadding)
+)
+
+// Credential used to be an interface that contained functions to encrypt, decrypt and authenticate.
```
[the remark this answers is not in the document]
jpcoenen: Fixed.

pkg/secrethub/credentials/encoding.go
```diff
@@ -0,0 +1,277 @@
...
+// We'll migrate away from using it and use smaller interfaces instead.
+// See Verifier, Decrypter and Encrypter for the smaller interfaces.
+type EncodableCredential interface {
```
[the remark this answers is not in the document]
jpcoenen: See comment below. Do we want to fix this now, as there is (almost?) no exposure to the user?

pkg/secrethub/credentials/encoding.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+// NewCredential is a shorthand function to decode a credential string and optionally
+// decrypt it with a passphrase. When an encrypted credential is given, the passphrase
+// cannot be empty.
+//
+// Note that when you want to customize the process of parsing and decoding/decrypting
+// a credential (e.g. to prompt only for a passphrase when the credential is encrypted),
+// it is recommended you use a CredentialParser instead (e.g. DefaultCredentialParser).
+func UnpackRSACredential(credential string, passphrase string) (*RSACredential, error) {
```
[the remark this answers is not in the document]
jpcoenen: Fixed

pkg/secrethub/http/options.go
jpcoenen: Renamed to `ClientOption`.

pkg/secrethub/credentials/providers.go
jpcoenen: Moved them. But not sure whether they are in the correct place now...

pkg/secrethub/client.go
SimonBarendse: https://github.com/secrethub/secrethub-cli/blob/v0.21.0/internals/secrethub/app.go#L81 used to accept a function that returns the interface, but now accepts a function that returns the struct (which does not implement the interface).

pkg/secrethub/client.go
SimonBarendse: Yes, should have been. With this change it becomes a problem though, as it is now accessible outside the package.

pkg/secrethub/credentials/readers.go
```diff
@@ -0,0 +1,83 @@
+package credentials
```
SimonBarendse: Is the idea that the CLI will implement a reader that will prompt?
This can be used as passphrase reader for `UseKey`.

pkg/secrethub/credentials/creators.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"errors"
+
+	awssdk "github.com/aws/aws-sdk-go/aws"
+
+	"github.com/secrethub/secrethub-go/internals/aws"
+	"github.com/secrethub/secrethub-go/internals/crypto"
+)
+
+type CreatedCredential interface {
```
jpcoenen: Not really satisfied with this. If anyone has a suggestion, let me know.

pkg/secrethub/credentials/providers.go
```diff
@@ -0,0 +1,116 @@
...
+	awssdk "github.com/aws/aws-sdk-go/aws"
+
+	"github.com/secrethub/secrethub-go/internals/auth"
+	"github.com/secrethub/secrethub-go/internals/aws"
+	"github.com/secrethub/secrethub-go/internals/crypto"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/credentials/sessions"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/internals/http"
+)
+
+type UsableCredential interface {
```
jpcoenen: Not really satisfied with this. If anyone has a suggestion, let me know.

pkg/secrethub/client.go
jpcoenen: Okay. That's clear. Not sure how big of a problem it is though. Do you have a suggestion of what else to call it?

pkg/secrethub/main_test.go
```diff
@@ -40,19 +42,20 @@ func init() {
...
+	handler.Handle("/v1/", http.StripPrefix("/v1", router))
 
 	// server is a test HTTP server used to provide mock API responses.
 	server := httptest.NewServer(handler)
 
-	opts := &ClientOptions{
-		ServerURL: server.URL,
+	opts := []ClientOption{
+		WithServerURL(server.URL),
+		WithCredentials(cred1),
```
SimonBarendse: Do we want to use the cred1 by default and overload it when we need a client for another account or do we want to be explicit about it in every test?

pkg/secrethub/service.go
```diff
@@ -3,34 +3,33 @@ package secrethub
...
 }
 
 type serviceService struct {
-	client client
+	client *Client
 }
 
 // Create creates a new service account for the given repo.
-func (s serviceService) Create(path string, description string, verifier Verifier, encrypter Encrypter) (*api.Service, error) {
+func (s serviceService) Create(path string, description string, credentialCreator credentials.Creator) (*api.Service, error) {
```
SimonBarendse: :ok_hand:

pkg/secrethub/main_test.go
SimonBarendse: Makes sense to use `cred1` by default, we use it everywhere :astonished:

pkg/secrethub/credentials/readers.go
jpcoenen: Oui oui. That's correct.

PR #97, opened 2019-08-13
 Create the credential for a service account in the client 

SimonBarendse:


pkg/secrethub/user.go
```diff
@@ -32,28 +32,38 @@ func (s userService) Me() (*api.User, error) {
...
 // Create creates a new user at SecretHub.
-func (s userService) Create(username, email, fullName string, verifier Verifier, encrypter Encrypter) (*api.User, error) {
+func (s userService) Create(username, email, fullName string) (*api.User, *RSACredential, error) {
 	err := api.ValidateUsername(username)
 	if err != nil {
-		return nil, errio.Error(err)
+		return nil, nil, err
 	}
 
 	err = api.ValidateEmail(email)
```
jpcoenen: Shouldn't this validation step happen in `create()`?

SimonBarendse: Why?

By validating upfront, we fail faster, before spending some time on generating the credential and account key.

jpcoenen: Shouldn't other callers of `create()` use the exact  same logic?

PR #99, opened 2019-08-14
Release v0.21.0

SimonBarendse:
### Added
- Add functionality to deal with complex password requirements
- Function to check whether an API error is caused by a resource not being found
- Use the new authentication header format to connect to the API server

### Changed
- Use narrower interfaces to pass the credential to the client
- Create the credential in the client when creating a new account (instead of passing a credential to the client)
- Pass permissions (read, write, admin) as strings instead of `api.Permission` to the client

### Removed
- Removed billing from the API
- Moved the server-side of authentication out of this repo


mackenbach:
When closing the release PR, could you comment why @SimonBarendse?

SimonBarendse:
@mackenbach We're waiting for some more changes to the way we initialize the client, before releasing this. That way, we reduce the number of breaking changes in the contract.

PR #101, opened 2019-08-15
Add JoinPaths function to join elements of a path

SimonBarendse:


internals/api/paths.go
```diff
@@ -557,3 +557,15 @@ func (n OrgName) Value() string {
...
+// JoinPaths joins any number of path elements into a single path.
+func JoinPaths(components ...string) string {
+	var processed []string
+	for _, c := range components {
+		trimmed := strings.Trim(c, pathSeparator)
+		if trimmed != "" {
+			processed = append(processed, trimmed)
+		}
+	}
+	return strings.Join(processed, pathSeparator)
```
mackenbach: This feels a bit like lazy code that works, but could just as easily be done by a simple for loop that trims the element and then appends it with a separator, e.g.:

```go
for i, c := range components {
		trimmed := strings.Trim(c, pathSeparator)
		if trimmed != "" && len(processed) == 0  {
			processed = append(processed, trimmed)
		} else if trimmed != "" {
                         processed = append(processed, pathSeparator, trimmed)
                }
	}
```

Not necessary to change it, but seems a bit silly to loop over the elements twice.

internals/api/paths_test.go
```diff
@@ -0,0 +1,46 @@
...
+func TestJoinPaths(t *testing.T) {
+	cases := map[string]struct {
+		elements []string
+		expected string
+	}{
+		"one path": {
+			elements: []string{"namespace/repo"},
+			expected: "namespace/repo",
+		},
+		"empty element": {
```
mackenbach: Consider the case for two empty elements?

internals/api/paths.go
SimonBarendse: As this is never called often nor with long arrays, the difference wouldn't be large.
Therefore I favor this somewhat simpler code.

SimonBarendse: Benchedmarked the difference, turned out the current proposal is actually faster (note that I've adapted the single loop to concat the strings though, as there would otherwise be two loops anyways).

```Go
// JoinPaths joins any number of path elements into a single path.
func JoinPaths(components ...string) string {
	var processed []string
	for _, c := range components {
		trimmed := strings.Trim(c, pathSeparator)
		if trimmed != "" {
			processed = append(processed, trimmed)
		}
	}
	return strings.Join(processed, pathSeparator)
}

func JoinPathsSingleLoop(components ...string) string {
	var processed string
	for _, c := range components {
		trimmed := strings.Trim(c, pathSeparator)
		if trimmed != "" && len(processed) == 0  {
			processed += trimmed
		} else if trimmed != "" {
			processed += pathSeparator + trimmed
		}
	}
	return processed
}
```

```
BenchmarkJoinPaths-8                    	 3000000	       347 ns/op
BenchmarkJoinPathsSingleLoop-8   	 5000000	       229 ns/op
```

SimonBarendse: Tried to make one more tiny optimization to the single loop version by using a `strings.Builder`. Wasn't better.

```Go
func JoinPathsSingleLoopStringBuilder(components ...string) string {
	var res strings.Builder
	for _, c := range components {
		trimmed := strings.Trim(c, pathSeparator)
		if trimmed != "" && res.Len() == 0  {
			res.WriteString(trimmed)
		} else if trimmed != "" {
			res.WriteString(pathSeparator)
			res.WriteString(trimmed)
		}
	}
	return res.String()
}
```

```
BenchmarkJoinPaths-8                          	 5000000	       322 ns/op
BenchmarkJoinPathsSingleLoop-8                	10000000	       202 ns/op
BenchmarkJoinPathsSingleLoopStringBuilder-8   	10000000	       221 ns/op
```

PR #104, opened 2019-08-15
Add pkg/secretpath

mackenbach:
This package implements utility functions for manipulating paths
compatible with SecretHub, e.g. namespaces, repositories, directories,
secrets and versions.

Had this one in the cooker for a while. Started it on a lazy sunday afternoon because I noticed we were doing quite some string manipulations in the Terraform provider and other projects that use the client library. 

The api.XXXPath types simply aren't usable enough so I tried to make something that is a bit more flexible. Inspired by Go's very own `filepath` [package](https://golang.org/pkg/path/filepath/).

### DIFF
--- a/pkg/secretpath/path.go
+++ b/pkg/secretpath/path.go
@@ -21,13 +21,6 @@ const (
 // Join joins any number of elements into a path, adding a separator
 // if necessary. Empty string elements are ignored.
 func Join(elements ...string) string {
-	switch len(elements) {
-	case 0:
-		return ""
-	case 1:
-		return strings.Trim(elements[0], elemSep)
-	}
-
 	result := ""
 	for _, e := range elements {
 		e = strings.Trim(e, elemSep)
@@ -38,7 +31,7 @@ func Join(elements ...string) string {
 		}
 	}
 
-	return result
+	return Clean(result)
 }
 
 // HasVersion returns true when a version suffix is specified in the path.
@@ -94,13 +87,13 @@ func AddVersion(path string, version int) string {
 		return path + suffix
 	}
 
-	return path[:i] + suffix
+	return Clean(path[:i] + suffix)
 }
 
 // Base returns the last element of a path. Trailing separators and version
 // numbers are removed.
 func Base(path string) string {
-	path = strings.Trim(path, elemSep)
+	path = Clean(path)
 
 	for i := len(path) - 1; i >= 0; i-- {
 		if path[i] == elemSepByte {
@@ -152,7 +145,7 @@ func Repo(path string) string {
 
 // Namespace returns the first element of a path, removing trailing separators.
 func Namespace(path string) string {
-	path = strings.Trim(path, elemSep)
+	path = Clean(path)
 	if len(path) == 0 {
 		return ""
 	}
@@ -172,6 +165,13 @@ func Namespace(path string) string {
 func Clean(path string) string {
 	if len(path) == 0 {
 		return ""
+	}
+
+	split := strings.SplitN(path, versionSep, 2)
+	path = split[0]
+	versionSuffix := ""
+	if len(split) > 1 {
+		versionSuffix = versionSep + split[1]
 	}
 
 	out := []byte{}
@@ -196,7 +196,7 @@ func Clean(path string) string {
 		out = append(out, path[i])
 	}
 
-	return string(out)
+	return string(out) + versionSuffix
 }
 
 // Count returns the number of elements in a path, excluding the version suffix.
--- a/pkg/secretpath/path.go
+++ b/pkg/secretpath/path.go
@@ -39,20 +39,17 @@ func Join(elements ...string) string {
 }
 
 // HasVersion returns true when a version suffix is specified in the path.
+// Note that even if the specified version suffix is invalid, this function 
+// still returns true.
 func HasVersion(path string) bool {
-	i := strings.LastIndex(path, versionSep)
-	return i >= 0 && i < len(path)-1
+	return strings.LastIndex(path, versionSep) >= 0  
 }
 
 // Version returns the version number suffix of a path, returning -1
-// when :latest or negative version numbers are given. If no version
-// suffix is set or it cannot be parsed to an integer, 0 is returned.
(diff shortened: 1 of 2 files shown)


mackenbach:
Still need to fix some logic. Added a few test cases that are failing for now.

mackenbach:
Finished the work, everything should be good now (aside from a few features I'd like to add at some point).

pkg/secretpath/path.go
```diff
@@ -0,0 +1,226 @@
...
+	versionSep     = string(versionSepByte)
+	latestSuffix   = ":latest"
+)
+
+// Join joins any number of elements into a path, adding a separator
+// if necessary. Empty string elements are ignored.
+func Join(elements ...string) string {
+	result := ""
+	for _, e := range elements {
+		e = strings.Trim(e, elemSep)
```
SimonBarendse: These checks and cleaning are redundant, as this will be cleaned by `Clean` as well.

```Go
result := ""
for _, e := range elements {
	result += elemSep + e
}

return Clean(result)
```
is sufficient.

---

If you do want to keep them in, they can be refactored like this to improve readability:

```Go
result := ""
for _, e := range elements {
	e = strings.Trim(e, elemSep)
	if len(e) == 0 {
		continue
	}
	if len(result) == 0 {
		result += e
	} else {
		result += elemSep + e
	}
}

return Clean(result)
```

pkg/secretpath/path.go
```diff
@@ -0,0 +1,226 @@
...
+	return Clean(result)
+}
+
+// HasVersion returns true when a version suffix is specified in the path.
+func HasVersion(path string) bool {
+	return hasVersion(path)
+}
+
+// hasVersion returns true when a version suffix is specified in the path.
+func hasVersion(path string) bool {
```
SimonBarendse: Why have this private func if there's a public function that does exactly the same: `HasVersion`?

pkg/secretpath/path.go
```diff
@@ -0,0 +1,226 @@
...
+	i := strings.LastIndex(path, versionSep)
+	return i >= 0 && i < len(path)-1
+}
+
+// Version returns the version number suffix of a path, returning -1
+// when :latest or negative version numbers are given. If no version
+// suffix is set or it cannot be parsed to an integer, 0 is returned.
+func Version(path string) int {
```
SimonBarendse: Should we use the same return for valid and invalid inputs?

We now have valid inputs:
`:latest` => -1
`` => 0

That return the same as these invalid inputs:
`:foo` => 0
`:-1` => -1

pkg/secretpath/path.go
```diff
@@ -0,0 +1,226 @@
...
+	if count == 1 {
+		return path
+	}
+	return ""
+}
+
+// Namespace returns the first element of a path, removing trailing separators.
+func Namespace(path string) string {
+	path = Clean(path)
+	if len(path) == 0 {
```
SimonBarendse: This check is redundant; For the empty string, the loop will have 0 iterations and the final return will be hit immediately.

pkg/secretpath/path.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	return path
+}
+
+// Clean returns the shortest path name equivalent to the given path
+// by lexical processing. It removes trailing and multiple separator
+// elements. Version suffixes are not removed.
+func Clean(path string) string {
+	if len(path) == 0 {
```
SimonBarendse: This check is redundant; For the empty string, `strings.SplitN` will return an array containing only the empty string; The loop will have 0 iterations, so the final return will be an empty byte array with the empty string concatenated.

pkg/secretpath/path.go
mackenbach: Ah, these used to be different. Now there not. You're right, the private function can be removed, moving the code to the public function.

pkg/secretpath/path.go
mackenbach: `-1` is a valid(ish) input, but in regards to your question: I think it should ignore the version when it's empty and it should ignore the version when it's not valid. So that would be the same case.

pkg/secretpath/path.go
mackenbach: It's not required, but a bit more performant I'd say. It's a small and very simple optimization to reduce the number of allocations. Seeing as the function get's called for almost every library call, it's worth the 'trade-off' IMO.

pkg/secretpath/path.go
SimonBarendse: > I think it should ignore the version when it's empty and it should ignore the version when it's not valid

What is `it` in this case?

---

If applications do not error on invalid versions, do invalid versions become valid input?

---

There are applications that want to treat empty versions and invalid versions differently. For example, a linter for run env files of the CLI will want to error on invalid versions, while accepting empty versions.

---

> -1 is a valid(ish) input

What do you mean with this? Should it be possible to supply `-1` as secret path version?  It's currently not accepted by `api.ValidateSecretPath`, so is not accepted by any of our applications (including the server).

pkg/secretpath/path.go
SimonBarendse: This check indeed reduces the number of memory allocations and CPU cycles needed when `Clean` is called with an empty path.

It also increases the CPU cycles needed when `Clean` is called with any other path.

And, as you said, the function gets called for almost every library call.

---

I'd say these library calls are useful when called on non-empty secret paths. Therefore, it might be better to optimize for this useful usage instead of an edge-case empty string usecase.

mackenbach:
@SimonBarendse I implemented all your comments, could you re-review?

pkg/secretpath/path.go
```diff
@@ -0,0 +1,215 @@
...
+		} else {
+			result += elemSep + e
+		}
+	}
+
+	return Clean(result)
+}
+
+// HasVersion returns true when a version suffix is specified in the path.
+// Note that even if the specified version suffix is invalid, this function
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
// Note that even if the specified version suffix is invalid, this function
```

PR #105, opened 2019-08-16
Add created_at field to api.Service

SimonBarendse:

PR #107, opened 2019-08-20
 Add type and meta_data fields to services 

SimonBarendse:
Merge #105 first

### DIFF
--- a/internals/api/credential.go
+++ b/internals/api/credential.go
@@ -44,8 +44,8 @@ type CredentialMetaDataKey string
 
 // Service metadata keys
 const (
-	ServiceMetaDataAWSKMSKey CredentialMetaDataKey = "kms_key"
-	ServiceMetaDataAWSRole   CredentialMetaDataKey = "role"
+	ServiceMetaDataAWSKMSKey CredentialMetaDataKey = "aws_kms_key"
+	ServiceMetaDataAWSRole   CredentialMetaDataKey = "aws_role"
 )
 
 // CreateCredentialRequest contains the fields to add a credential to an account.
--- a/internals/api/credential.go
+++ b/internals/api/credential.go
@@ -40,12 +40,21 @@ func (a CredentialType) Validate() error {
 	return ErrInvalidAlgorithm
 }
 
+type CredentialMetaDataKey string
+
+// Service metadata keys
+const (
+	ServiceMetaDataAWSKMSKey CredentialMetaDataKey = "kms_key"
+	ServiceMetaDataAWSRole   CredentialMetaDataKey = "role"
+)
+
 // CreateCredentialRequest contains the fields to add a credential to an account.
 type CreateCredentialRequest struct {
-	Type        CredentialType `json:"type"`
-	Fingerprint string         `json:"fingerprint"`
-	Name        string         `json:"name,omitempty"`
-	Verifier    []byte         `json:"verifier"`
+	Type        CredentialType    `json:"type"`
+	Fingerprint string            `json:"fingerprint"`
+	Name        string            `json:"name,omitempty"`
+	Verifier    []byte            `json:"verifier"`
+	Metadata    map[string]string `json:"metadata,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


internals/api/service.go
```diff
@@ -23,13 +23,32 @@ var (
...
-	Description string     `json:"description"`
-	CreatedBy   *uuid.UUID `json:"created_by,omitempty"`
+	AccountID   *uuid.UUID        `json:"account_id"`
+	ServiceID   string            `json:"service_id"`
+	Repo        *Repo             `json:"repo"`
+	Description string            `json:"description"`
+	CreatedBy   *uuid.UUID        `json:"created_by,omitempty"`
+	CreatedAt   time.Time         `json:"created_at"`
+	Type        ServiceType       `json:"type"`
+	MetaData    map[string]string `json:"meta_data,omitempty"`
```
mackenbach: In my head it's `metadata` without a space instead of `meta data`. It also seems to be what other APIs use as name, e.g. Stripe API.

internals/api/service.go
```diff
@@ -23,13 +23,32 @@ var (
 	)
 )
 
+type ServiceType string
+
+// Service types
+const (
+	ServiceTypeAWS ServiceType = "aws"
+	ServiceTypeRSA ServiceType = "rsa"
```
mackenbach: @jpcoenen based on our discussions today, should this be `rsa` or `key`?

internals/api/service.go
```diff
@@ -23,13 +23,32 @@ var (
...
+)
+
+type ServiceMetaDataKey string
+
+// Service metadata keys
+const (
+	ServiceMetaDataAWSKMSKey ServiceMetaDataKey = "kms_key"
+	ServiceMetaDataAWSRole   ServiceMetaDataKey = "role"
```
mackenbach: Should this be `role`, `role_arn`, `arn` or `aws_role`?

SimonBarendse: I think `role` suffices.
ARN is the format in which the role is given. We typically don't describe the format in the name.
We can name it `aws_role`. Might be a bit redundant, as the type is already `aws`. I'm fine with both.

mackenbach: I think it's best to distinguish between type specific fields. `aws_role` or `role_arn` implies that it's on AWS. Yes the type is already of `aws`, but I can imagine other types also using the `role` key. Then I'd rather be explicit. Small thing though.

SimonBarendse: Oke, let's make it `aws_role` and `aws_kms_key` then :+1:

internals/api/service.go
```diff
@@ -23,13 +23,23 @@ var (
...
+)
+
 // Service represents a service account on SecretHub.
 type Service struct {
-	AccountID   *uuid.UUID `json:"account_id"`
-	ServiceID   string     `json:"service_id"`
-	Repo        *Repo      `json:"repo"`
-	Description string     `json:"description"`
-	CreatedBy   *uuid.UUID `json:"created_by,omitempty"`
+	AccountID   *uuid.UUID        `json:"account_id"`
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

internals/api/credential.go
```diff
@@ -40,12 +40,21 @@ func (a CredentialType) Validate() error {
 	return ErrInvalidAlgorithm
 }
 
+type CredentialMetaDataKey string
+
+// Service metadata keys
+const (
+	ServiceMetaDataAWSKMSKey CredentialMetaDataKey = "aws_kms_key"
+	ServiceMetaDataAWSRole   CredentialMetaDataKey = "aws_role"
```
mackenbach: There are more occurences of `MetaData` that should be `Metadata`

SimonBarendse:
Replaced with #109 and #110

SimonBarendse:
Replaced with #109 and #110

PR #109, opened 2019-08-20
Add credential to api.Service

SimonBarendse:


internals/api/service.go
```diff
@@ -24,13 +24,22 @@ var (
 	ErrAccessDeniedToKMSKey = errAPI.Code("access_denied").StatusError("access to KMS key is denied", http.StatusForbidden)
 )
 
+type ServiceType string
+
+// Service types
+const (
+	ServiceTypeAWS ServiceType = "aws"
+	ServiceTypeRSA ServiceType = "rsa"
```
jpcoenen: @mackenbach and I just discussed that it makes more sense to expose this to the user as type `key`. Then the user is not bothered with _what_ type of key. The actual key type used could be stored in the metadata, if we wish.

jpcoenen: @mackenbach and I just discussed that it might make more sense to expose this to the user as type `key`. Then the user is not bothered with _what_ type of key. The actual key type used could be stored in the metadata if we wish. But for basic use, the fact that the RSA algorithm is used, is too specific.

Do you agree?

SimonBarendse: I agree that the "rsa" might be to specific to display to the user in e.g. the CLI.
Is it also to specific to return in the API response?
If we want to be able to make specific assumptions about what metadata will be present based on the service type, it might be good to be specific here.

I think we should distinguish between the type for display purposes and the type used for logic.
We can translate the more specific type used for the application logic to a type that can be used for display purposes.

SimonBarendse: Discussed this. We want to embed api.Credential in the api.Service response instead.

PR #112, opened 2019-08-23
Add setting HTTP User-Agent with application specific info

jpcoenen:


pkg/secrethub/client.go
```diff
@@ -8,6 +8,10 @@ import (
 	"github.com/secrethub/secrethub-go/pkg/secrethub/internals/http"
 )
 
+const (
+	UserAgentPrefix = "SecretHub/v1 GoClient/" + ClientVersion
```
SimonBarendse: Should this be public?

pkg/secrethub/client.go
```diff
@@ -42,6 +46,27 @@ type Client struct {
 	// repoindexKeys are the keys used to generate blind names in the repo.
 	// These are cached
 	repoIndexKeys map[api.RepoPath]*crypto.SymmetricKey
+
+	appInfo *AppInfo
+}
+
+// AppInfo contains information about the application that is using the SecretHub client.
+// It is used to identify the application to the SecretHub API.
```
mackenbach: Do we need users of the library to set this themselves? If so, consider adding a comment like: 

`It is used to identify the application to the SecretHub API, helping us a lot when analyzing errors and usage. So please provide this information when integrating with this libary.`

pkg/secrethub/client.go
```diff
@@ -42,6 +46,27 @@ type Client struct {
...
+type AppInfo struct {
+	Name            string
+	Version         string
+	OperatingSystem string
```
mackenbach: Do we want to add a `URL` field with a url of the project? Would help us a bunch too I guess.

pkg/secrethub/client.go
mackenbach: In the future, we could also (perhaps) expose an analytics route that allows integration builders to see how actively their 'App' is being used, e.g. call `api.secrethub.io/v1/analytics/apps/secrethub-cli` responds with e.g. the number of API calls made for that app in the last 30 days. 

Not a priority, at all, but something cool to consider

pkg/secrethub/client.go
```diff
@@ -8,6 +8,10 @@ import (
 	"github.com/secrethub/secrethub-go/pkg/secrethub/internals/http"
 )
 
+const (
+	userAgentPrefix = "SecretHub/v1 GoClient/" + ClientVersion
```
mackenbach: Why do we say `GoClient/vX.Y.Z` here, but do we say `secrethub-go` elsewhere as `DefaultUserAgent = "secrethub-go"`?

pkg/secrethub/client.go
mackenbach: Also, do we want to add a utility function or something to help auto-detect the OS? And, what info do we want to include in the OS? Found this library that can do quite a lot it seems: https://github.com/matishsiao/goInfo

mackenbach: Or even better yet, the Docker team has invented this wheel already:

- Detect OS: https://github.com/moby/moby/blob/master/pkg/parsers/operatingsystem/operatingsystem_linux.go
- Detect Kernel: https://github.com/moby/moby/tree/master/pkg/parsers/kernel
- Pack version information into UserAgent header format: https://github.com/moby/moby/tree/master/pkg/useragent

PR #113, opened 2019-08-26
 Send metadata when creating AWS credentials

SimonBarendse:


internals/api/credential.go
```diff
@@ -33,6 +33,8 @@ var (
 // CredentialMetadataKey is a key that can be used for the metadata of a credential.
 type CredentialMetadataKey string
 
+type CredentialMetadata map[CredentialMetadataKey]string
```
mackenbach: Do we really need the type `CredentialMetadataKey`? With the `CredentialMetadata` already being typed is `type CredentialMetadata map[string]string` sufficient?

SimonBarendse: By typing the key, I'm trying to make more explicit that only a subset of all possible keys is allowed to be used: the keys defined below.

SimonBarendse: I've added a comment to make it extra clear.

PR #116, opened 2019-08-27
Always supply full ARNs in credential metadata

SimonBarendse:


internals/aws/service_creator.go
```diff
@@ -31,28 +31,36 @@ type CredentialCreator struct {
 
 // NewCredentialCreator returns a CredentialCreator that uses the provided AWS KMS key and IAM role to create a new credential.
 // The AWS credential is configured with the optionally provided aws.Config.
-func NewCredentialCreator(keyID, role string, cfgs ...*aws.Config) (*CredentialCreator, error) {
+func NewCredentialCreator(keyID, role string, cfgs ...*aws.Config) (*CredentialCreator, map[string]string, error) {
```
mackenbach: It looks like this signature is not compatible with the changes made in #117

internals/aws/service_creator.go
```diff
@@ -114,20 +122,17 @@ func GetEncryptRequest(plaintext string, keyID string, kmsSvc kmsiface.KMSAPI) (
...
-func parseRole(role string, stsSvc stsiface.STSAPI) (string, error) {
+// to find and return the corresponding full ARN. The given role can be:
+// - A role name (e.g. my-role)
+// - A role name, prefixed with "role/" (e.g. role/my-role)
+// - A full ARN (e.g. arn:aws:iam::123456789012:role/my-role)
+//
+// Note that this is a best effort attempt so the returned value is not guaranteed to
+// always be a valid ARN.
+func parseRole(role string, accountID string) string {
 	if strings.Contains(role, ":") {
```
mackenbach: This is fairly fuzzy matching logic to check if the given string is already a valid role ARN, correct? Do we want to use something a bit more sophisticated here or not?

SimonBarendse: This hasn't been changed.

As also documented in the comment, the function does only return valid output on valid input. This is oke for its use-case, as the given metadata is only stored when creating the service succeeds, which only happens when the role and key are valid.

internals/aws/service_creator.go
SimonBarendse: We can evaluate the role and key in `CreateAWS` instead.

PR #117, opened 2019-08-27
AWS polish

jpcoenen:
Final changes to make the new client usable with the CLI.

### DIFF
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -17,14 +17,23 @@ const (
 
 // ClientAdapter is an interface that can be used to consume the SecretHub client and is implemented by secrethub.Client.
 type ClientAdapter interface {
+	// AccessRules returns a service used to manage access rules.
 	AccessRules() AccessRuleService
+	// Accounts returns a service used to manage SecretHub accounts.
 	Accounts() AccountService
+	// Dirs returns a service used to manage directories.
 	Dirs() DirService
+	// Me returns a service used to manage the current authenticated account.
 	Me() MeService
+	// Orgs returns a service used to manage shared organization workspaces.
 	Orgs() OrgService
+	// Repos returns a service used to manage repositories.
 	Repos() RepoService
+	// Secrets returns a service used to manage secrets.
 	Secrets() SecretService
+	// Services returns a service used to manage non-human service accounts.
 	Services() ServiceService
+	// Users returns a service used to manage (human) user accounts.
 	Users() UserService
 }
 
@@ -136,47 +145,47 @@ func Must(c *Client, err error) *Client {
 	return c
 }
 
-// AccessRules returns an AccessRuleService.
+// AccessRules returns a service used to manage access rules.
 func (c *Client) AccessRules() AccessRuleService {
 	return newAccessRuleService(c)
 }
 
-// Accounts returns an AccountService.
+// Accounts returns a service used to manage SecretHub accounts.
 func (c *Client) Accounts() AccountService {
 	return newAccountService(c)
 }
 
-// Dirs returns an DirService.
+// Dirs returns a service used to manage directories.
 func (c *Client) Dirs() DirService {
 	return newDirService(c)
 }
 
-// Me returns a MeService.
+// Me returns a service used to manage the current authenticated account.
 func (c *Client) Me() MeService {
 	return newMeService(c)
 }
 
-// Orgs returns an OrgService.
+// Orgs returns a service used to manage shared organization workspaces.
 func (c *Client) Orgs() OrgService {
 	return newOrgService(c)
 }
 
-// Repos returns an RepoService.
+// Repos returns a service used to manage repositories.
 func (c *Client) Repos() RepoService {
 	return newRepoService(c)
 }
 
-// Secrets returns an SecretService.
+// Secrets returns a service used to manage secrets.
 func (c *Client) Secrets() SecretService {
 	return newSecretService(c)
 }
 
-// Services returns an ServiceService.
+// Services returns a service used to manage non-human service accounts.
 func (c *Client) Services() ServiceService {
 	return newServiceService(c)
 }
 
-// Users returns an UserService.
+// Users returns a service used to manage (human) user accounts.
 func (c *Client) Users() UserService {
 	return newUserService(c)
 }
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -93,7 +93,6 @@ func NewClient(with ...ClientOption) (*Client, error) {
 	client := &Client{
 		httpClient:    http.NewClient(),
 		repoIndexKeys: make(map[api.RepoPath]*crypto.SymmetricKey),
-		ConfigDir:     &configdir.Dir{},
 	}
 	err := client.with(with...)
 	if err != nil {
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -190,6 +190,9 @@ func (c *Client) Users() UserService {
 	return newUserService(c)
 }
 
+// DefaultCredential returns a reader pointing to the configured credential,
+// sourcing it either from the SECRETHUB_CREDENTIAL environment variable or
+// from the configuration directory.
 func (c *Client) DefaultCredential() credentials.Reader {
 	envCredential := os.Getenv("SECRETHUB_CREDENTIAL")
 	if envCredential != "" {
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -1,6 +1,8 @@
 package secrethub
 
 import (
+	"os"
+
 	"github.com/secrethub/secrethub-go/internals/api"
 	"github.com/secrethub/secrethub-go/internals/crypto"
 	"github.com/secrethub/secrethub-go/internals/errio"
@@ -102,7 +104,7 @@ func NewClient(with ...ClientOption) (*Client, error) {
 
 	// Try to use default key credentials if none provided explicitly
 	if client.decrypter == nil {
-		err := WithCredentials(credentials.UseKey(nil, nil))(client)
+		err := WithCredentials(credentials.UseKey(client.DefaultCredential()))(client)
 		// nolint: staticcheck
 		if err != nil {
 			// TODO: log that default credential was not loaded.
@@ -177,4 +179,13 @@ func (c *Client) Services() ServiceService {
 // Users returns an UserService.
 func (c *Client) Users() UserService {
 	return newUserService(c)
+}
+
+func (c *Client) DefaultCredential() credentials.Reader {
+	envCredential := os.Getenv("SECRETHUB_CREDENTIAL")
+	if envCredential != "" {
+		return credentials.FromString(envCredential)
+	}
+
+	return c.ConfigDir.Credential()
 }
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -4,6 +4,7 @@ import (
 	"github.com/secrethub/secrethub-go/internals/api"
 	"github.com/secrethub/secrethub-go/internals/crypto"
 	"github.com/secrethub/secrethub-go/internals/errio"
+	"github.com/secrethub/secrethub-go/pkg/secrethub/configdir"
 	"github.com/secrethub/secrethub-go/pkg/secrethub/credentials"
 	"github.com/secrethub/secrethub-go/pkg/secrethub/internals/http"
 )
@@ -47,7 +48,8 @@ type Client struct {
 	// These are cached
 	repoIndexKeys map[api.RepoPath]*crypto.SymmetricKey
 
-	appInfo *AppInfo
+	appInfo   *AppInfo
+	ConfigDir *configdir.Dir
 }
 
 // AppInfo contains information about the application that is using the SecretHub client.
@@ -80,12 +82,22 @@ func NewClient(with ...ClientOption) (*Client, error) {
 	client := &Client{
 		httpClient:    http.NewClient(),
 		repoIndexKeys: make(map[api.RepoPath]*crypto.SymmetricKey),
+		ConfigDir:     &configdir.Dir{},
 	}
 	for _, option := range with {
 		err := option(client)
 		if err != nil {
 			return nil, err
 		}
+	}
+
+	// ConfigDir should be fully initialized before loading any default credentials.
+	if client.ConfigDir == nil {
+		configDir, err := configdir.Default()
+		if err != nil {
+			return nil, err
+		}
+		client.ConfigDir = configDir
 	}
 
 	// Try to use default key credentials if none provided explicitly
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -95,11 +95,9 @@ func NewClient(with ...ClientOption) (*Client, error) {
 		repoIndexKeys: make(map[api.RepoPath]*crypto.SymmetricKey),
 		ConfigDir:     &configdir.Dir{},
 	}
-	for _, option := range with {
-		err := option(client)
-		if err != nil {
-			return nil, err
-		}
+	err := client.with(with...)
+	if err != nil {
+		return nil, err
 	}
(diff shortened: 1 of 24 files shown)


pkg/secrethub/configdir/dir.go
```diff
@@ -0,0 +1,71 @@
...
+		Path: filepath.Join(c.Path, "credential"),
+	}
+}
+
+type CredentialFile struct {
+	Path string
+}
+
+func (f *CredentialFile) Write(p []byte) error {
+	// TOOD: correct permission?
```
golangcibot: `TOOD` is a misspelling of `TODO` (from `misspell`)

pkg/secrethub/configdir/dir.go
```diff
@@ -0,0 +1,71 @@
...
+	err := os.MkdirAll(filepath.Dir(f.Path), 0600)
+	if err != nil {
+		return err
+	}
+	// TOOD: correct permission?
```
golangcibot: `TOOD` is a misspelling of `TODO` (from `misspell`)

pkg/secrethub/configdir/dir.go
```diff
@@ -0,0 +1,71 @@
...
+	if _, err := os.Stat(f.Path); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+func (f *CredentialFile) Read() ([]byte, error) {
+	file, err := os.Open(f.Path)
+	if os.IsNotExist(err) {
+		// TOOD: return more usable error
```
golangcibot: `TOOD` is a misspelling of `TODO` (from `misspell`)

pkg/secrethub/client_options.go
```diff
@@ -47,6 +48,13 @@ func WithAppInfo(appInfo *AppInfo) ClientOption {
 	}
 }
 
+func WithConfigDir(configDir configdir.Dir) ClientOption {
```
mackenbach: missing comments

pkg/secrethub/configdir/dir.go
```diff
@@ -0,0 +1,71 @@
+package configdir
```
mackenbach: This package misses documentation. I can write some.

pkg/secrethub/configdir/dir.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/mitchellh/go-homedir"
+)
+
+type Dir struct {
+	Path string
+}
+
+func Default() (*Dir, error) {
+	// TODO: move to constant?
```
mackenbach: As long as we use it in just one place, I'm fine with using a magical string as it actually improves readability. If we use the same thing multiple times though, we should move it to a constant.

pkg/secrethub/client.go
```diff
@@ -80,6 +84,7 @@ func NewClient(with ...ClientOption) (*Client, error) {
 	client := &Client{
 		httpClient:    http.NewClient(),
 		repoIndexKeys: make(map[api.RepoPath]*crypto.SymmetricKey),
+		ConfigDir:     &configdir.Dir{},
```
SimonBarendse: Should this be initialized to `nil`? So that the nil check on line 96 can set it?

pkg/secrethub/configdir/dir.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/mitchellh/go-homedir"
+)
+
+type Dir struct {
+	Path string
```
SimonBarendse: Should this be publicly adjustable?

pkg/secrethub/configdir/dir.go
```diff
@@ -0,0 +1,83 @@
+// Package configdir provides simple functions to manage the SecretHub
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

pkg/secrethub/configdir/dir.go
SimonBarendse: i.e. should it be exported?

pkg/secrethub/configdir/dir.go
mackenbach: It used to be:

```go
	// defaultCredentialFileMode is the filemode to assign to the credential file.
	defaultCredentialFileMode = os.FileMode(0600)
	// defaultProfileDirFileMode is the filemode to assign to the configuration directory.
	defaultProfileDirFileMode = os.FileMode(0700)
```

So if you don't have a strong preference we can simply use those values?

pkg/secrethub/credentials/creators.go
```diff
@@ -31,13 +34,36 @@ func CreateKey() *KeyCreator {
...
+	encrypter Encrypter
+}
+
+func (ac *awsCreator) Create() error {
+	creator, err := aws.NewCredentialCreator(ac.kmsKeyID, ac.roleARN, ac.awsCfg...)
+	if err != nil {
+		return err
+	}
+	ac.encrypter = creator
+	ac.verifier = creator
```
mackenbach: why does this have two fields that are the same? Can we then have one field that are returned for both the `Verifier` and `Authenticator` functions?

pkg/secrethub/credentials/creators.go
```diff
@@ -31,13 +34,36 @@ func CreateKey() *KeyCreator {
...
-	return c.key, c.key, map[string]string{}, nil
+	kc.key = key
+	return nil
+}
+
+func (kc *KeyCreator) Metadata() map[string]string {
+	return map[string]string{}
+}
+
+type awsCreator struct {
```
mackenbach: This type is stuck in between multiple functions of the `KeyCreator`. Let's move  it somehwere else in the file for clarity.

pkg/secrethub/credentials/key.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (k Key) Encrypter() Encrypter {
+	return k.key
+}
+
+func (k Key) Provide(httpClient *http.Client) (auth.Authenticator, Decrypter, error) {
+	return k.key, k.key, nil
+}
+
+func (k Key) Passphrase(passphraseReader Reader) Key {
```
mackenbach: What does this do? And this type is missing godoc as well.

pkg/secrethub/credentials/key.go
```diff
@@ -0,0 +1,104 @@
...
+		passBasedKey, err := NewPassBasedKey(passphrase)
+		if err != nil {
+			return nil, err
+		}
+		return EncodeEncryptedCredential(k.key, passBasedKey)
+	}
+	return EncodeCredential(k.key)
+}
+
+func ImportKey(credentialReader Reader, passphraseReader Reader) (Key, error) {
```
mackenbach: Is there a reason for having both `ImportKey` and `readKey`? They both seem to do the exact same.

mackenbach: And let's add some godoc here that explains also the retry mechanism.

pkg/secrethub/credentials/key.go
```diff
@@ -0,0 +1,104 @@
...
+				continue
+			} else if err != nil {
+				return Key{}, err
+			}
+
+			return Key{key: credential}, nil
+		}
+
+		return Key{}, ErrCannotDecryptCredential
+	}
```
mackenbach: This is a medium complex function. Adding a bit more whitespace or comments might help. Took me a while to figure out that the `for` loop actually contains a return so this is the non-encrypted branch of the statement.

pkg/secrethub/credentials/providers.go
```diff
@@ -47,51 +42,30 @@ func UseAWS(awsCfg ...*awssdk.Config) Provider {
...
-			}
-			passphrase, err := ioutil.ReadAll(passReader)
-			if err != nil {
-				return nil, nil, err
-			}
-			key, err := NewPassBasedKey(passphrase)
-			if err != nil {
-				return nil, nil, err
-			}
+type KeyProvider struct {
```
mackenbach: This stuff definitely needs some godocs

pkg/secrethub/credentials/providers.go
```diff
@@ -47,51 +42,30 @@ func UseAWS(awsCfg ...*awssdk.Config) Provider {
 //   3. The credential file found in <user's home directory>/.secrethub/credential.
```
mackenbach: Is this comment still up to date?

pkg/secrethub/credentials/readers.go
```diff
@@ -1,83 +1,50 @@
...
-	"github.com/mitchellh/go-homedir"
 )
 
-// Errors
-var (
-	ErrCannotFindHomeDir = errCredentials.Code("cannot_find_home_dir").ErrorPref(
-		"cannot find your home directory: %s",
-	)
-)
+type Reader interface {
```
mackenbach: Needs godoc

pkg/secrethub/client.go
```diff
@@ -102,7 +104,7 @@ func NewClient(with ...ClientOption) (*Client, error) {
 
 	// Try to use default key credentials if none provided explicitly
 	if client.decrypter == nil {
-		err := WithCredentials(credentials.UseKey(nil, nil))(client)
+		err := WithCredentials(credentials.UseKey(client.DefaultCredential()))(client)
```
mackenbach: This is a bit of a 'hacky' fix to apply an option after the client has already been initialized, correct?

Should we then create a dedicated (private) `with(options ...Options)` function for this?

pkg/secrethub/client.go
```diff
@@ -102,7 +104,7 @@ func NewClient(with ...ClientOption) (*Client, error) {
...
 		// nolint: staticcheck
 		if err != nil {
 			// TODO: log that default credential was not loaded.
```
mackenbach: TODO found

pkg/secrethub/credentials/encoding.go
```diff
@@ -168,8 +168,8 @@ func newParser(decoders []Decoder) parser {
 }
 
 // parse parses a credential string.
-func (p parser) parse(raw string) (*encodedCredential, error) {
-	parts := strings.Split(raw, ".")
+func (p parser) parse(raw []byte) (*encodedCredential, error) {
+	parts := strings.Split(string(raw), ".")
```
mackenbach: If you're doing this, you can also just use `bytes.Split` and avoid the uncessary casts here.

pkg/secrethub/configdir/dir.go
mackenbach: > Should this be publicly adjustable?

Yes that is explicitly a feature. When the CLI is passed the `--config-dir` flag, you need to be able to create a config dir with a given path. 

Whether that should be with a `New(path string)` function, `Set(path string)` function or just my modifying the field is a style choice.

SimonBarendse: Ahh, from the current `Dir` type, I expected a one-on-one mapping between a directory on your file system and a struct, but if I understand you correctly, the struct only contains the configuration of which directory to use and can be switched to point to a different directory?

To prevent similar confusion, could we change the name to reflect this? e.g. `Config`

Alternatively, we could use the `New(path string)` to create a new `Dir` instance when using a different directory.

pkg/secrethub/configdir/dir.go
jpcoenen: That's fine by me

pkg/secrethub/credentials/encoding.go
jpcoenen: Did not know about this function. Let's use it.

pkg/secrethub/client.go
jpcoenen: Hmm, that's an interesting approach.

pkg/secrethub/credentials/creators.go
jpcoenen: :check:

pkg/secrethub/credentials/creators.go
jpcoenen: Is it though? I moved the `CreateAWS()` function above this. Is it clearer now?

pkg/secrethub/credentials/key.go
jpcoenen: It is the same, now, yes. So we can change it.

pkg/secrethub/client.go
jpcoenen: Tried it and works nicely.

pkg/secrethub/credentials/providers.go
jpcoenen: Nope. Fixed.

pkg/secrethub/credentials/key.go
jpcoenen: Rearranged it a little. Better?

pkg/secrethub/credentials/key.go
jpcoenen: Done so.

pkg/secrethub/credentials/key.go
jpcoenen: Added a comnent

pkg/secrethub/credentials/providers.go
jpcoenen: Added

pkg/secrethub/client.go
jpcoenen: Is this what we want to pick up now though? We do not have any logging in place on the client at the moment.

mackenbach: I don't know, just spotted the TODO and thought I'd bring it up. Your call what to do here.

pkg/secrethub/credentials/creators.go
mackenbach: AH cool, maybe I misread. Looks good. 👍

pkg/secrethub/credentials/key.go
```diff
@@ -65,9 +65,17 @@ func ImportKey(credentialReader, passphraseReader Reader) (Key, error) {
...
 
+		// Try up to three times to get the correct passphrase.
 		for i := 0; i < 3; i++ {
-			credential, err := decryptKey(passphraseReader, encoded)
+			passphrase, err := passphraseReader.Read()
+			if err != nil {
+				return Key{}, err
+			}
+			if len(passphrase) == 0 {
+				return Key{}, errors.New("no passphrase given")
```
mackenbach: It used to retry when you don't provide a passphrase, now it exits. Is that an intentional change in behavior?

pkg/secrethub/client.go
SimonBarendse: We decided not to do it right now.

SimonBarendse: #119

pkg/secrethub/configdir/dir.go
mackenbach: Yes, the types in the `configdir.Dir` and `configdir.CredentialFile` merely point to a location on the filesystem and offer utility methods for working with those files/dirs. 

To me the name `configdir.Dir` makes sense, it's the directory where configuration is stored. Later we may add a `configdir.ConfigFile` or `configdir.SettingsFile` with configuration. The `configdir.XXXFile` types are stored within the `configdir.Dir` type. 

Now the `New(path string)` thingy could actually work. May even be a bit nicer because you're reducing the surface area of contact of the package. Not suuuper necessary now, but a nice improvement I'd say.

@jpcoenen your call what to do.

jpcoenen: @SimonBarendse: have a look at how I solved this.

pkg/secrethub/credentials/key.go
jpcoenen: It was not an intentional change, but it is intentional behaviour. Because I did not see much purpose in trying again if you do not enter anything.

But we discussed it here and conclude that it might be a bit more user-friendly to retry.

jpcoenen: Added, see https://github.com/secrethub/secrethub-go/pull/117/commits/737803b2fb291293bafd01aa0e1a338fef4893ab

PR #118, opened 2019-08-28
Add Fingerprint to the Verifier interface and implement it for aws creds

SimonBarendse:


pkg/secrethub/credentials/credentials.go
```diff
@@ -11,6 +11,8 @@ type Verifier interface {
 	Type() api.CredentialType
 	// AddProof adds the proof of this credential's possession to a CreateCredentialRequest.
 	AddProof(req *api.CreateCredentialRequest) error
+	// Fingerprint returns the key identifier by which the server can identify the verifier.
+	Fingerprint() (string, error)
```
mackenbach: Question: what made you decide to expand the interface while it looks like getting the fingerprint can already be achieved with the current interface (using `Verifier` and `Type` functions)?

From your code above:

```go
verifier, err := c.Verifier()
if err != nil {
	return "", err
}

 return api.GetFingerprint(c.Type(), verifier)
```

Maybe I'm missing something.

SimonBarendse:
#120 replaces this PR

PR #127, opened 2019-09-03
Add function to validate uuid strings

SimonBarendse:


internals/api/uuid/uuid.go
```diff
@@ -39,3 +47,18 @@ func (u *UUID) IsZero() bool {
...
 }
+
+// Validate validates a uuid string is a valid UUID.
+func Validate(str string) error {
+	_, err := FromString(str)
+	return err
+}
+
+// IsErrInvalidUUID returns whether the given error is returned for an invalid uuid.
+func IsErrInvalidUUID(err error) bool {
```
jpcoenen: Wouldn't it make more sense to have a more general `apiErrorEqual(err1, err2 error) bool` that checks whether the namespace and code of an API error are equal? Then we can reuse that in other cases.

internals/errio/errors_test.go
```diff
@@ -112,3 +114,40 @@ func TestUnexpectedStatusError(t *testing.T) {
...
 }
+
+func TestEquals(t *testing.T) {
+	cases := map[string]struct {
+		a        PublicError
+		b        error
+		expected bool
+	}{
+		"completely equal errors": {
+			a: Namespace("app").Code("invalid foo").Error("invalid value for foo: value should be a string"),
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

PR #128, opened 2019-09-05
Use values instead of pointers for uuid values that are always set

SimonBarendse:
On responses from the server, the UUID values are always given
(except for the parent directory of a dir), therefore, we can return
the values themselves instead of pointers. By doing so, the consuming
applications don't have to dereference the pointers everywhere they
are used.

This change becomes more relevant now, as we are creating endpoints
to get resources by their IDs, so the IDs will be consumed more often
in the future.


internals/api/dirfs.go
```diff
@@ -149,18 +149,18 @@ func (t Tree) AbsSecretPath(secretID *uuid.UUID) (*SecretPath, error) {
...
 	}
 
-	dir, ok := t.Dirs[*dirID]
+	dir, ok := t.Dirs[dirID]
 	if !ok {
 		return "", ErrDirNotFound
 	}
 
-	parentPath, err := t.AbsDirPath(dir.ParentID)
+	parentPath, err := t.AbsDirPath(*dir.ParentID)
```
mackenbach: This seems like we can avoid the cast here by looking at the signature of either the type or the function?

internals/api/dirfs_test.go
```diff
@@ -23,21 +23,21 @@ func TestAbsDirPath(t *testing.T) {
...
 		Name:     "dir",
-		ParentID: repoDir.DirID,
+		ParentID: &repoDir.DirID,
 	}
 
 	subdir := &Dir{
 		DirID:    uuid.New(),
 		Name:     "subdir",
-		ParentID: dir.DirID,
+		ParentID: &dir.DirID,
```
mackenbach: Why is this still a pointer?

internals/api/repo_test.go
```diff
@@ -51,7 +51,7 @@ func TestCreateRepoRequest_Validate(t *testing.T) {
 func TestInviteUserRequest_Validate_Success(t *testing.T) {
 
 	inviteRequest := &api.InviteUserRequest{
-		AccountID:  accountIDUser1,
+		AccountID:  &accountIDUser1,
```
mackenbach: Is this field optional? If not, should it be a pointer?

internals/api/secret_test.go
```diff
@@ -96,16 +96,16 @@ func TestCreateSecretRequest_Validate_Unique(t *testing.T) {
 				EncryptedData: testCiphertextAES,
 
 				EncryptedNames: []api.EncryptedNameRequest{{
-					AccountID:     accountID,
+					AccountID:     &accountID,
```
mackenbach: Same here, is this field optional and should it be a pointer?

internals/api/secret_test.go
```diff
@@ -136,12 +136,12 @@ func TestCreateSecretRequest_Validate_EncryptedNameAndKeyForEachAccount(t *testi
 		EncryptedData: testCiphertextAES,
 
 		EncryptedNames: []api.EncryptedNameRequest{{
-			AccountID:     uuid.New(),
+			AccountID:     UUID(uuid.New()),
```
mackenbach: Why do we need this cast? Doesn't the `uuid` package return the right type?

internals/api/secret_version_test.go
```diff
@@ -44,8 +44,9 @@ func TestCreateSecretVersionRequest_Validate_MaxSize(t *testing.T) {
 			t.Fatal(err)
 		}
 
+		id := uuid.New()
 		r := CreateSecretVersionRequest{
-			SecretKeyID:   uuid.New(),
+			SecretKeyID:   &id,
```
mackenbach: Same here, is the `SecretKeyID` field optional and should it therefore be a pointer?

internals/api/uuid_test.go
```diff
@@ -0,0 +1,10 @@
+package api_test
+
+import (
+	"github.com/secrethub/secrethub-go/internals/api/uuid"
+)
+
+// UUID returns the pointer to the given uuid.UUID value.
+func UUID(id uuid.UUID) *uuid.UUID {
```
mackenbach: What is the purpose of having this function and does it have its own file?

pkg/secrethub/crypto.go
```diff
@@ -49,10 +49,10 @@ func (c *Client) encryptSecretFor(secret *api.Secret, accounts ...*api.Account)
 
 		encryptedName := api.EncryptedNameForNodeRequest{
 			EncryptedNameRequest: api.EncryptedNameRequest{
-				AccountID:     account.AccountID,
+				AccountID:     &account.AccountID,
 				EncryptedName: encryptedSecretName,
 			},
-			NodeID: secret.SecretID,
+			NodeID: &secret.SecretID,
```
mackenbach: Optional field or required?

pkg/secrethub/repo_user.go
```diff
@@ -58,7 +58,7 @@ func (s repoUserService) Invite(path string, username string) (*api.RepoMember,
 	}
 
 	in := &api.InviteUserRequest{
-		AccountID:  account.AccountID,
+		AccountID:  &account.AccountID,
```
mackenbach: Optional field or required?

pkg/secrethub/secret_version.go
```diff
@@ -167,7 +167,7 @@ func (c *Client) createSecretVersion(secretPath api.SecretPath, data []byte, sec
 
 	in := &api.CreateSecretVersionRequest{
 		EncryptedData: encryptedData,
-		SecretKeyID:   secretKey.SecretKeyID,
+		SecretKeyID:   &secretKey.SecretKeyID,
```
mackenbach: Optional field or required?

pkg/secrethub/secret_key.go
```diff
@@ -57,7 +57,7 @@ func (c *Client) createSecretKey(secretPath api.SecretPath) (*api.SecretKey, err
 		}
 
 		encryptedFor[i] = api.EncryptedKeyRequest{
-			AccountID:    account.AccountID,
+			AccountID:    &account.AccountID,
```
mackenbach: Optional field or required?

pkg/secrethub/crypto.go
```diff
@@ -63,8 +63,8 @@ func (c *Client) encryptSecretFor(secret *api.Secret, accounts ...*api.Account)
 			}
 
 			encryptedKeys[keyIndex] = api.SecretKeyMemberRequest{
-				AccountID:    account.AccountID,
-				SecretKeyID:  decryptedKey.SecretKeyID,
+				AccountID:    &account.AccountID,
```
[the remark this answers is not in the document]
SimonBarendse: I have only changed to values instead of pointers for responses (of http- and regular client). For http requests, I have left the pointers, so that we are able to check whether the value was set or left to the zero value. In this specific case, we do so in `api.SecretKeyMemberRequest.Validate`.

internals/api/dirfs_test.go
SimonBarendse: Because the parent ID is optional. It is not set for root directories.

internals/api/uuid_test.go
SimonBarendse: It's for test purposes only, because we create a lot of requests there with fake UUIDs. We could also refactor it to call `uuid.New` itself and rename it to something like `fakeUUID`. Would that make things clearer?

internals/api/dirfs.go
SimonBarendse: What do you mean by that? `t.AbsDirPath` traverses the tree to get the full path of the given directory. It's not a type cast.

pkg/secrethub/crypto.go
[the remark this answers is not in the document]
mackenbach: Alright, but is that really what we should check? If it is a required field, can we check if it is the zero value instead?

internals/api/dirfs.go
mackenbach: Ah I said `cast` but I meant `pointer dereference`, brain fart I guess.

internals/api/uuid_test.go
mackenbach: Why do we need it in the first place? Why not simply call `uuid.New` in the places where you generate a random UUID for test purposes?

internals/api/dirfs.go
SimonBarendse: The pointer dereference is made because the parent id field is optional. It is not set for root directories. As we've checked whether the directory is the root directory on the first line of this function, we can safely dereference here.

internals/api/uuid_test.go
SimonBarendse: Because `uuid.New` is used server-side, were we'd also consume the value and not the pointer, as in (almost?) all cases the id is a required field.

PR #129, opened 2019-09-06
Use AWS identity provider with SECRETHUB_IDENTITY_PROVIDER=aws envvar

SimonBarendse:


pkg/secrethub/client.go
```diff
@@ -110,7 +115,19 @@ func NewClient(with ...ClientOption) (*Client, error) {
 
 	// Try to use default key credentials if none provided explicitly
 	if client.decrypter == nil {
-		err := client.with(WithCredentials(credentials.UseKey(client.DefaultCredential())))
+		identityProvider := os.Getenv("SECRETHUB_IDENTITY_PROVIDER")
+
+		var provider credentials.Provider
+		switch identityProvider {
+		case "key", "":
```
mackenbach: Should this be case insensitive?

mackenbach: Also after looking at the code for a while I now see you have the `""` empty string match here too. That's not clear at a glance. Looking at this switch it looks like the default is to error. Can we make this a bit more readable?

pkg/secrethub/client.go
```diff
@@ -110,7 +115,19 @@ func NewClient(with ...ClientOption) (*Client, error) {
...
+			provider = credentials.UseKey(client.DefaultCredential())
+		case "aws":
+			provider = credentials.UseAWS()
+		default:
+			return nil, ErrUnknownIdentityProvider(identityProvider)
```
mackenbach: Should it error here or just ignore it and use the default (a key) instead?

SimonBarendse: If the identity provider is explicitly set to an invalid value, I think it's best to error.

For example, when making a typo in the envvar, setting e.g. `SECRETHUB_IDENTITY_PROVIDER=awss`, I think it would be better to error and explain that `awss` is not a valid value than to continue and either:
- Find a credential at `~/.secrethub/credential` and authenticate as another user (which could result in permission errors)
- Not find a credential and make unauthenticated requests (which will return an error on routes that require authentication)
I think in both cases, the returned errors don't explain the issue (typo in envvar) clearly.

pkg/secrethub/client.go
SimonBarendse: > Should this be case insensitive?

Yes, good suggestion. I'll also submit a PR to [secrethub/secrethub-cli](https://github.com/secrethub/secrethub-cli) to make the envvar and flag case insensitive there as well.

SimonBarendse: secrethub/secrethub-cli#160

SimonBarendse: > 

Also after looking at the code for a while I now see you have the "" empty string match here too. That's not clear at a glance. Looking at this switch it looks like the default is to error. Can we make this a bit more readable?

I've swapped the `""` and `"key"` values to make the empty string stand out more. Does this help?

PR #133, opened 2019-09-13
Add Read and ReadString convenience functions

SimonBarendse:
This makes it easier to use the client directly in applications.


pkg/secrethub/secret.go
```diff
@@ -19,6 +19,11 @@ type SecretService interface {
 	// Versions returns a SecretVersionService.
 	Versions() SecretVersionService
 
+	// Read gets a secret version, with sensitive data decrypted.
```
mackenbach: Consider adding in the comment here something like: `for more extensive handling of secret versions, see the Versions() function`

PR #134, opened 2019-09-16
Add/improve godoc package comments

mackenbach:


internals/api/docs.go
```diff
@@ -0,0 +1,3 @@
+// Package api provides request and response types for
+// interacting with the SecretHub API.
```
SimonBarendse: Currently, I think the `api` package contains a lot of things (which we should separate one step at a time):
- HTTP request and response types
- client response types (some of which are the same types as for the http response)
- sorters for arrays of client response types (e.g. `SortAccessRules`)
- helper functions that operate on client response types (e.g. `GetFingerprint`)
- types used for validated input (e.g. `AccountName`, these are no longer used for input of client functions, but are used in the response types)

SimonBarendse: Most of these should be moved to either `secrethub` or `secrethub/internal/http` package.

mackenbach: Yes agreed. Most of the types should be moved to their respective package that uses/exposes them. This is a bit of a medium/longer term goal though. Let's see when we can start moving towards this.

PR #138, opened 2019-10-15
Update README examples to use new constructor and ReadString

SimonBarendse:


README.md
```diff
@@ -53,26 +53,57 @@ For details on all functionality of this library, see the [GoDoc][godoc] documen
...
+    "github.com/secrethub/secrethub-go/pkg/secrethub"
+)
 
-// Generate a slice of 32 alphanumeric characters.
-data, err := randchar.NewGenerator(false).Generate(32)
-secret, err = client.Secrets().Write("path/to/secret", data)
+func main() {
+    client, _ := secrethub.NewClient()
+    rand, _ := randchar.NewRand(randchar.Alphanumeric)
+    data, _ := rand.Generate(30)
```
mackenbach: Can we simplify this even further to take the default rand? E.g. I'd like to see utility functions on the package that use a `randchar.Default` to make super easy one-liners e.g. `randchar.Generate(30)`. Alternatively, `randchar.Default.Generate(30` also works.

README.md
```diff
@@ -53,26 +53,57 @@ For details on all functionality of this library, see the [GoDoc][godoc] documen
...
+    _, _ = client.Secrets().Write("path/to/secret", data)
+}
 ```
 
 > **Note:** to use the SecretHub Go client, you need to provide a credential for your __SecretHub__ account.
```
mackenbach: This note should actually be at the `getting started` part. Also, is it `# Installation` or `# Usage`? You're not really installing something, you're using a library.

README.md
```diff
@@ -53,26 +53,57 @@ For details on all functionality of this library, see the [GoDoc][godoc] documen
```
mackenbach: The heading structure of the document is a bit off:

```
## Getting Started
<no content>
### Installation
<content>
## Examples
<content>
### Read a secret
<content>
### Write a secret
<content>
### Generate a secret
<content>
```

I'd suggest making it something like:

```
## Usage (or getting started)
<`go get` instructions>
<import (AND client construct instructions?)>
<notice about getting an account (which should include that `secrethub.NewClient` auto-loads the credential)>

### Examples

#### Read a secret
<code>
#### Write a secret
<code>
#### Generate a random secret
<code>
```

README.md
mackenbach: Because the underlying OS random number generator can return an error. We could make it so it doesn't return an error and panics instead, but I don't like that so much.

mackenbach:
@SimonBarendse I see you've created PR #139 for the `getting help` suggestion, could you share your plans/thoughts for the other suggestions I made?

README.md
mackenbach: Ah gotcha, yes NewRand could not return an error, but options currently return an error. There are some valid cases where it should, I think, e.g. when you add mutually exclusive options

README.md
SimonBarendse: https://github.com/secrethub/secrethub-go/pull/141

README.md
SimonBarendse: https://github.com/secrethub/secrethub-go/pull/141

SimonBarendse:
@mackenbach I've opened https://github.com/secrethub/secrethub-go/pull/141 to follow up on your other suggestions.

README.md
SimonBarendse: We can update the example when we have this shortcut default functionality.

SimonBarendse: https://github.com/secrethub/secrethub-go/pull/142

PR #141, opened 2019-10-17
Restructure the readme

SimonBarendse:


README.md
```diff
@@ -47,13 +45,21 @@ import (
 > **Note:** only packages inside the `/pkg` directory should be considered library code that you can use in your projects. 
 > All other code is not guaranteed to be backwards compatible and may change in the future.
 
-## Examples
+### Credential
```
mackenbach: I know I said to move it here, but now it feels like it makes more sense to move it below the examples section. As a user, I'm more eager to see the code than the caveats.

PR #143, opened 2019-10-22
Only add query params to request when not empty

SimonBarendse:
Don't add a query parameter when no subject types are given.

Request `/v1/namespaces/<workspace>/repos/<repo>/events` instead of `/v1/namespaces/<workspace>/repos/<repo>/events?subject_types=`.

### DIFF
--- a/pkg/secrethub/internals/http/client.go
+++ b/pkg/secrethub/internals/http/client.go
@@ -249,8 +249,8 @@ func (c *Client) AuditRepo(namespace, repoName string, subjectTypes api.AuditSub
 	out := []*api.Audit{}
 	rawURL := fmt.Sprintf(pathRepoEvents, c.base, namespace, repoName)
 	query := make(url.Values)
-	for _, subjectType := range subjectTypes {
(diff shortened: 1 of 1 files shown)


pkg/secrethub/internals/http/client.go
```diff
@@ -247,7 +247,14 @@ func (c *Client) DeleteRepo(namespace, repoName string) error {
...
 	out := []*api.Audit{}
-	rawURL := fmt.Sprintf(pathRepoEvents+"?subject_types=%s", c.base, namespace, repoName, subjectTypes.Join(","))
+	rawURL := fmt.Sprintf(pathRepoEvents, c.base, namespace, repoName)
+	query := make(url.Values)
+	if len(subjectTypes) > 0 {
+		query.Set("subject_types", subjectTypes.Join(","))
+	}
+	if len(query) > 0 {
+		rawURL += "?" + query.Encode()
+	}
```
mackenbach: I'm assuming you're planning to add more query params later so that's why you made it with two dependent `if` statements instead of everything in one?

PR #144, opened 2019-10-23
Add EventIterator funcs

SimonBarendse:
These iterators make it possible to iterate over an infinite stream of audit events. The events are fetched in batches (pages) from the API, but this is abstracted away in the iterator.

Changes:
- Add `Repos().EventIterator` and `Secrets().EventIterator`
- Validate server URL when option is applied
- Depcreate `Repos().ListEvents` and `Secrets().ListEvents`

Depcrecated `ListEvents` functions, use `EventIterator` instead:

`client.Repos().ListEvents(path, nil)` => `client.Repos().EventIterator(path)`

`client.Repos().ListEvents(path, subjectTypes)` => `client.Repos().ListEvents(path, secrethub.OnlySubjectTypes(subjectTypes))`

`client.Secrets().ListEvents(path, nil)` => `client.Secrets().EventIterator(path)`

`client.Secrets().ListEvents(path, subjectTypes)` => `client.Secrets().ListEvents(path, secrethub.OnlySubjectTypes(subjectTypes))`


pkg/secrethub/internals/http/client.go
```diff
@@ -245,11 +245,40 @@ func (c *Client) DeleteRepo(namespace, repoName string) error {
 }
 
 // AuditRepo gets the audit events for a given repo.
-func (c *Client) AuditRepo(namespace, repoName string, subjectTypes api.AuditSubjectTypeList) ([]*api.Audit, error) {
-	out := []*api.Audit{}
-	rawURL := fmt.Sprintf(pathRepoEvents+"?subject_types=%s", c.base, namespace, repoName, subjectTypes.Join(","))
-	err := c.get(rawURL, true, &out)
-	return out, errio.Error(err)
+func (c *Client) AuditRepo(namespace, repoName string, subjectTypes api.AuditSubjectTypeList) (*Paginator, error) {
```
SimonBarendse: Now that we're changing this signature, let's also change the way we pass filter options such as the subject types, so that we don't change the signature again when adding more filter options.

mackenbach: Let's discuss first whether we actually want to break the signature. Alternatively we could e.g. add a new function and deprecate the old one.

pkg/secrethub/internals/http/client.go
```diff
@@ -245,11 +245,40 @@ func (c *Client) DeleteRepo(namespace, repoName string) error {
...
+}
+
+func (c *Client) auditPage(url url.URL) *Paginator {
+	events := make([]api.Audit, 50)
+	return NewPaginator(
+		url,
+		&events,
+		func(vs interface{}) string {
+			events := *vs.(*[]api.Audit)
+			return events[len(events)-1].EventID.String()
```
mackenbach: What happens when `len(events)` is `0`? And is that a case we need to account for?

pkg/secrethub/internals/http/client.go
```diff
@@ -245,11 +245,40 @@ func (c *Client) DeleteRepo(namespace, repoName string) error {
...
+
+	return c.auditPage(*requestURL), nil
+}
+
+func (c *Client) auditPage(url url.URL) *Paginator {
+	events := make([]api.Audit, 50)
+	return NewPaginator(
+		url,
+		&events,
+		func(vs interface{}) string {
```
mackenbach: wut is a `vs`??? Can we improve the name to auto-document itself?

pkg/secrethub/internals/http/pagination.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"net/url"
+)
+
+func NewPaginator(url url.URL, target interface{}, getStartingAfter func(interface{}) string, toItems func(interface{}) []interface{}, client *Client) *Paginator {
+	return &Paginator{
+		url:              &url,
+		target:           target,
+		getStartingAfter: getStartingAfter,
```
mackenbach: Do we also support `getStartingBefore` or `getEndingBefore`? If so, how would that fit?

pkg/secrethub/internals/http/client.go
SimonBarendse: This function should never be called on an empty list. This is ensured by this check:
https://github.com/secrethub/secrethub-go/pull/144/files/83e130dc31279abcb00411bf94e53acb412c5f2f#diff-cc52c786e26f396dead3234863b06611R34

Alternatively, we could make the function return an error and only add the request parameter when the function returns no error (`nil`). That would mean we have to add this zero length check for every type of paginator though (auditevents, users etc.), instead of defining it once in the paginator.

When we add the link header to the response, we can drop this function altogether, as we can then get the next page from the response.

pkg/secrethub/internals/http/pagination.go
SimonBarendse: No, that is not yet supported with this PR.

When we add functionality for that, I imagine doing so by injecting the next page behavior (strategy pattern) in the paginator, but we're flexible to choose whatever we want when we get to it. We can safely change this constructor, as it is only used internally (hence the `internals` package).

For exposing this functionality trough the `ListEvents` functions to the user, see: https://github.com/secrethub/secrethub-go/pull/144/files/83e130dc31279abcb00411bf94e53acb412c5f2f#r338147963

pkg/secrethub/internals/http/client.go
mackenbach: Then how about `values` 😄

pkg/secrethub/internals/http/pagination.go
mackenbach: I think you misunderstood my intention. I'm asking: does the API have this functionality and therefore should we add that functionality to the client as well?

pkg/secrethub/internals/http/client.go
SimonBarendse: Good question :+1:
I think this also depends on how we add pagination on the API. For audit events we have done so on the existing route and by doing so changed the behavior of the `ListEvents` function. It now limits it's results to the first 50 events.
Because of this change in behavior, I feel more for dropping the function altogether.

For other routes, we can deprecate first before removing the list functions. We can already add the functions that return iterators before adding pagination on those routes (we just iterate over one page that contains all entries). We can then remove the function before adding pagination:
add iterator funcs => deprecate list funcs => :hourglass_flowing_sand: => remove list funcs => add pagination on routes

Therefore, I propose to change the function name of the audit iterators regardless of whether we want the list functions to keep the same signature. So that we can follow the same naming pattern for the other routes.

How about `Iterator` for a function name?
e.g. `client.Users().Members().Iterator()` vs `client.Users().Members().List()`
`client.Users().Repo().EventIterator()` vs `client.Users().Repo().ListEvents()`

I like that it signals that nothing happens yet. `List` is active, it does something (retrieves the events from the API), `Iterator` just returns something with which you can do something (iterating).

pkg/secrethub/internals/http/pagination.go
SimonBarendse: Yes, the API has this functionality. Yes, let's add it to the client as well. Let's do that in a separate PR though, to keep changes clear and concise and be able to release functionality sooner.

pkg/secrethub/iterator.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// Errors
+var (
+	IteratorDone = errio.Namespace("iterator").Code("done").Error("there are no more items left")
+)
+
+type iterator struct {
+	pag   *http.Paginator
+	i     int
```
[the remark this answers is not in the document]
SimonBarendse: Yes, let's rename this to `currentIndex` :+1:

pkg/secrethub/audit.go
```diff
@@ -32,3 +32,21 @@ func (c *Client) decryptAuditEvents(events ...*api.Audit) error {
 
 	return nil
 }
+
+type AuditEventIterator struct {
+	*iterator
```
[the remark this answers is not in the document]
SimonBarendse: No, so I changed it to embed without a pointer.

pkg/secrethub/internals/http/client.go
```diff
@@ -245,18 +245,40 @@ func (c *Client) DeleteRepo(namespace, repoName string) error {
...
+			events := *vs.(*[]api.Audit)
```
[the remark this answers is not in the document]
SimonBarendse: You cannot index a pointer to a list (`events[<index>]`)

pkg/secrethub/internals/http/client.go
```diff
@@ -245,18 +245,40 @@ func (c *Client) DeleteRepo(namespace, repoName string) error {
...
+			return events[len(events)-1].EventID.String()
+		},
+		func(vs interface{}) []interface{} {
+			events := *vs.(*[]api.Audit)
```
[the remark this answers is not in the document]
SimonBarendse: You cannot call `len()` on a pointer to a list.

pkg/secrethub/internals/http/client.go
```diff
@@ -245,18 +245,40 @@ func (c *Client) DeleteRepo(namespace, repoName string) error {
...
+		q.Set("subject_types", subjectTypes.Join(","))
+		requestURL.RawQuery = q.Encode()
 	}
-	err := c.get(rawURL, true, &out)
-	return out, errio.Error(err)
+
+	return c.auditPage(*requestURL), nil
+}
+
+func (c *Client) auditPage(url url.URL) *Paginator {
```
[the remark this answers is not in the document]
SimonBarendse: I'd love to move away from `url.Parse`. Instead, we could validate the URL (and return an error) when setting the base URL of the client.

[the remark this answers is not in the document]
SimonBarendse: Now that I've said this, let's do this now, so that the signature of `client.Repos().EventIterator(..)` changes to only return a `Paginator` and no `error`.

pkg/secrethub/internals/http/pagination.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+type Paginator struct {
+	url              *url.URL
+	target           interface{}
+	getStartingAfter func(interface{}) string
+	toItems          func(interface{}) []interface{}
+
+	client *Client
```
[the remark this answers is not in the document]
SimonBarendse: I've implemented 2

pkg/secrethub/internals/http/client.go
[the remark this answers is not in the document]
jpcoenen: Didn't know that :open_mouth:

pkg/secrethub/internals/http/client.go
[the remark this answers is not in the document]
jpcoenen: The more you know.

pkg/secrethub/internals/http/client.go
SimonBarendse: These checks are now moved together.

pkg/secrethub/audit.go
```diff
@@ -32,3 +32,21 @@ func (c *Client) decryptAuditEvents(events ...*api.Audit) error {
...
+	c *Client
```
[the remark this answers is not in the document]
SimonBarendse: https://github.com/secrethub/secrethub-go/pull/144#discussion_r338454016

pkg/secrethub/internals/http/client.go
```diff
@@ -208,54 +207,96 @@ func (c *Client) GetUser(username string) (*api.User, error) {
...
+	}
+
+	return newAuditPaginator(requestURL, c)
+}
+
+func newAuditPaginator(requestURL url.URL, client *Client) *AuditPaginator {
+	return &AuditPaginator{
+		fetchPage: func(target *[]api.Audit, query url.Values) error {
+			q := requestURL.Query()
+			for k, _ := range query {
```
golangcibot: should omit 2nd value from range; this loop is equivalent to `for k := range ...` (from `golint`)

pkg/secrethub/audit.go
```diff
@@ -32,3 +33,32 @@ func (c *Client) decryptAuditEvents(events ...*api.Audit) error {
...
+		},
+		paginator:          paginator,
+		decryptAuditEvents: client.decryptAuditEvents,
+	}
+}
+
+type AuditEventIterator struct {
+	iterator
+	decryptAuditEvents func(...*api.Audit) error
+	paginator *http.AuditPaginator
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	paginator          *http.AuditPaginator
```

pkg/secrethub/repo.go
```diff
@@ -16,6 +16,23 @@ type RepoService interface {
...
+	// 	   event, err := events.Next()
+	//	   if err == secrethub.IteratorDone {
+	//	       break
+	//	   }
+	//	   if err != nil {
+	//	       // Handle error
+	//	   }
+	//
+	//     // Use event
+	// }
```
jpcoenen: Code has to be intended to be printed as code in the godoc.

pkg/secrethub/audit.go
```diff
@@ -62,3 +62,11 @@ func (it *AuditEventIterator) Next() (api.Audit, error) {
...
 	return audit, nil
 }
+
+type AuditEventIterationOption func(*AuditEventIterator) error
+
+func OnlySubjectTypes(subjectTypes api.AuditSubjectTypeList) AuditEventIterationOption {
+	return func(it *AuditEventIterator) error {
+		return http.OnlySubjectTypes(subjectTypes)(it.paginator)
+	}
+}
```
jpcoenen: Do I correctly understand that this means that we have to duplicate any option in both the `http` and the `secrethub` package?

pkg/secrethub/repo.go
jpcoenen: You already fixed this.

pkg/secrethub/audit.go
SimonBarendse: Yes, options on the paginator are passed to the `http` package (similarly as we do for client options).

Options on the iterator are of course not passed to the `http` package.

pkg/secrethub/repo.go
```diff
@@ -14,11 +14,32 @@ type RepoService interface {
...
 	Get(path string) (*api.Repo, error)
+	// EventIterator returns an iterator that retrieves all audit events for a given repo.
+	//
+	// Usage:
+	//  it, err := client.Repos().EventIterator(path)
+	//  if err != nil {
+	//  	   // Handle error
+	//  }
+	//  for {
+	//      event, err := events.Next()
```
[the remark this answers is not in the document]
SimonBarendse: 1. It validates the supplied repo path (as do other client functions that accept a path)
2. Yes :+1: 
3. Sounds good :+1:

pkg/secrethub/audit.go
```diff
@@ -32,3 +33,40 @@ func (c *Client) decryptAuditEvents(events ...*api.Audit) error {
...
+	decryptAuditEvents func(...*api.Audit) error
+	paginator          *http.AuditPaginator
+}
+
+func (it *AuditEventIterator) Next() (api.Audit, error) {
+	item, err := it.next()
+	if err != nil {
+		return api.Audit{}, err
+	}
+	audit := item.(api.Audit)
```
mackenbach: Do we need to check these type casts?

pkg/secrethub/iterator.go
```diff
@@ -0,0 +1,40 @@
...
+// Errors
+var (
+	IteratorDone = errio.Namespace("iterator").Code("done").Error("there are no more items left")
+)
+
+type paginator interface {
+	Next() ([]interface{}, error)
+}
+
+type iterator struct {
```
mackenbach: One thing to be careful about with these types of abstractions is how this works if multiple goroutines call `iterator.Next()`. Seeing as we're exposing this as a library, I feel like the iterator should be able to handle that nicely, e.g. with an internal mutex?

What are your thoughts on this subject?

pkg/secrethub/audit.go
```diff
@@ -32,3 +33,40 @@ func (c *Client) decryptAuditEvents(events ...*api.Audit) error {
...
+	err = it.decryptAuditEvents(&audit)
```
mackenbach: Question: we've been noticing the client side decryption of audit events can take `1-3s`  (whereas the server response is `~100ms`). The flow used to be: fetch a page, decrypt the page, show the page. Now it looks like the flow will be: fetch a page, wait til `iterator.Next()` and for each call decrypt that event.

What are your thoughts on the impact on latency of this iterator pattern? Fixing the latency is probably a separate issue and worthy of its own PR, but if this seriously aggravates the latency issue, then we should make a conscious decision about that now.

SimonBarendse: The flow used to be:
1. fetch events
2. sequentially decrypt all events
3. return all events
---
Now it is:
1. fetch page of events
2. decrypt 1 event
3. return 1 event
4. repeat at 2 (or 1 when end of page is reached)

So, when `Next` is called sequentially from 1 go-routine, decryption of events is done sequentially in that go-routine (status quo maintained).

When `Next` is called concurrently, decryption of events is also done concurrently. This reduces the latency problem when multiple cores are available.

We can improve as follows:
1. fetch page of events + concurrently decrypt events
2. wait for 1 event to be decrypted
3. return event
4. repeat at 2 (or 1 when end of page is reached (or when we've buffering implemented, when the buffer is smaller than desired))

Now decryption is always done concurrently, independent of how the caller uses the iterator.

pkg/secrethub/audit.go
SimonBarendse: `newAuditEventIterator` (line 37) only accepts `*http.AuditPaginator` which always return audit events. As long as the `AuditEventIterator` is created from that constructor, the type check is not necessary.

`AuditEventIterator`s created outside the `secrethub` package, cannot set any of its fields. Does it make sense to build in error checks for such usage?
I believe that would panic on a nil-pointer dereference on a call to next of the `iterator`. The zero-value for the iterator has zero-values for all of its fields. This would mean a value of `nil` for the paginator. Calling `iterator.next` calls `iterator.paginator.Next` which is a nil-pointer reference.

pkg/secrethub/audit_test.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"testing"
+
+	"github.com/secrethub/secrethub-go/internals/api"
+	"github.com/secrethub/secrethub-go/internals/api/uuid"
+	"github.com/secrethub/secrethub-go/internals/assert"
+)
+
+type fakeAuditPaginator struct {
+	events []api.Audit
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	events   []api.Audit
```

pkg/secrethub/repo.go
```diff
@@ -107,6 +130,45 @@ func (s repoService) ListEvents(path string, subjectTypes api.AuditSubjectTypeLi
...
+	res := newAuditEventIterator(paginator, s.client)
+
+	for _, option := range options {
+		err := option(&res)
+		if err != nil {
+			return AuditEventIterator{}, err
+		}
+	}
+
+	return res, nil
```
golangcibot: copylocks: return copies lock value: github.com/secrethub/secrethub-go/pkg/secrethub.AuditEventIterator contains github.com/secrethub/secrethub-go/pkg/secrethub.iterator contains sync.Mutex (from `govet`)

pkg/secrethub/secret.go
```diff
@@ -198,6 +221,49 @@ func (s secretService) ListEvents(path string, subjectTypes api.AuditSubjectType
...
+	res := newAuditEventIterator(paginator, s.client)
+
+	for _, option := range options {
+		err := option(&res)
+		if err != nil {
+			return AuditEventIterator{}, err
+		}
+	}
+
+	return res, nil
```
golangcibot: copylocks: return copies lock value: github.com/secrethub/secrethub-go/pkg/secrethub.AuditEventIterator contains github.com/secrethub/secrethub-go/pkg/secrethub.iterator contains sync.Mutex (from `govet`)

pkg/secrethub/repo.go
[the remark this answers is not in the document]
SimonBarendse: Update 2 and 3.

For 1, turns out the `EventIterator` also returns errors that might result from evaluating the options that are passed (e.g. from validation of the option, supplying conflicting options, supplying an option twice etc.).
I think it would be confusing to move those errors to the first call of `Next` as well. What do you think?

pkg/secrethub/iterator.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+func (it *iterator) next() (interface{}, error) {
+	it.mutex.Lock()
+	if it.items == nil || (len(it.items) > 0 && len(it.items) <= it.currentIndex) {
+		var err error
+		it.items, err = it.paginator.Next()
+		if err != nil {
+			it.mutex.Unlock()
```
mackenbach: I'm not really seeing why a `defer it.mutex.Unlock()` wouldn't do the same trick here. Your commit message mentions something about a recursive call to `it.next`, but I do not see it.

pkg/secrethub/repo.go
```diff
@@ -107,6 +130,45 @@ func (s repoService) ListEvents(path string, subjectTypes api.AuditSubjectTypeLi
...
+// EventIterator returns an iterator that retrieves all audit events for a given repo.
+//
+// Usage:
+//  iter, err := client.Repos().EventIterator(path)
+//  if err != nil {
+//  	   // Handle error
+//  }
+//  for {
+//      event, err := iter.Next()
+//	    if err == secrethub.IteratorDone {
```
mackenbach: This indentation is off.

pkg/secrethub/repo.go
```diff
@@ -107,6 +130,45 @@ func (s repoService) ListEvents(path string, subjectTypes api.AuditSubjectTypeLi
...
+//	        break
+//	    }
+//	    if err != nil {
```
mackenbach: Personally I feel this should be an `else if`, but I know you guys may feel differently about that.

pkg/secrethub/repo.go
```diff
@@ -107,6 +130,45 @@ func (s repoService) ListEvents(path string, subjectTypes api.AuditSubjectTypeLi
...
+//	        // Handle error
```
mackenbach: let's just make this say `return err` or something.

pkg/secrethub/secret.go
```diff
@@ -13,7 +13,28 @@ type SecretService interface {
...
+	// 	    event, err := iter.Next()
+	//	    if err == secrethub.IteratorDone {
+	//	        break
+	//	    }
+	//	    if err != nil {
+	//	        // Handle error
+	//	    }
+	//
+	//      // Use event
+	//  }
```
mackenbach: Personally I don't feel the usage has become any clearer by our abstraction. We now have three error checks that don't all make super much sense intuitively:

- On `.EventIterator`
- At `if err == secrethub.IteratorDone` (and why is that in the `secrethub` package?)
- At `if err != nil`

Not entirely sure what is causing the awkwardness, but it feels to me like one of these cases would be simpler:

```go
iter, err := client.Secrets().EventIterator(path)
for {
   event, err := iter.Next()
   if err == iterator.Done {
      break
   } else if err != nil {
      // Handle error
   }
   // Use event
}
```
Or Stripe flavored:
```go
iter, err := client.Secrets().EventIterator(path)
for iter.HasNext() {
   event := iter.Next()
   // Use event
}
err := iter.Err()
if err != nil {
   // handle error
}
```

pkg/secrethub/iterator.go
SimonBarendse: The recursive call is on line 45

pkg/secrethub/repo.go
SimonBarendse: Why? It's up to the user to decide how to deal with the error, isn't it?

pkg/secrethub/secret.go
SimonBarendse: https://github.com/secrethub/secrethub-go/pull/144#discussion_r340206520

pkg/secrethub/iterator.go
mackenbach: Ah dot it, totally missed that one. Could this benefit from a small wrapper function that does the mutex stuff and then calls this function so we don't have to deal with all those mutexes here?

mackenbach:
Let's discuss in person tomorrow, that will go way faster.

jpcoenen:
One thing I notice that I see a lot of things added to the `secrethub` namespace, but what does an `Iterator` have to do with `secrethub`?

pkg/secrethub/iterator.go
SimonBarendse: Created `nextUnsafe` so that the name makes clear that the function shouldn't be used directly. `next` obtains and releases the lock and calls `nextUnsafe`.

mackenbach:
Nice addition with `nextUnsafe` @SimonBarendse 😍

pkg/secrethub/repo.go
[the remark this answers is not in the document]
SimonBarendse: Updated to construct the paginator on first call of `Next`, so that the construction of the iterator does not return an error.

pkg/secrethub/internals/http/client.go
```diff
@@ -208,61 +207,106 @@ func (c *Client) GetUser(username string) (*api.User, error) {
...
+		q := pag.requestURL.Query()
+		q.Set("starting_after", events[len(events)-1].EventID.String())
+		pag.requestURL.RawQuery = q.Encode()
+	}
+
+	res := make([]interface{}, len(events))
+	for i, event := range events {
+		res[i] = event
+	}
+	return res, nil
```
jpcoenen: I'm not a big fan of what happens here. For example, it does not look very extendable to other types. But I am okay with leaving it as is and refactoring it when we implement other paginators. Especially since it is internal code anyway.

SimonBarendse: I believe my idea when implementing this was about the same; When we go for another paginator, I'd like to have the link header, so that all paginators determine the next page URL in the same way. Then, there's more that's the same for all paginators and we can extract that all together.

SimonBarendse:
> Looks good now :+1: One question left: do you think it is tested thorougly enough?

Yes. Next to the unit test of the audit event iterator, I've added e2e tests for both the repo and secret audit (these are not in this repo). Together, this gives me enough confidence in the implementation.

pkg/secrethub/fakeclient/repo.go
```diff
@@ -9,15 +9,18 @@ import (
...
-	Creater        RepoCreater
-	Deleter        RepoDeleter
-	Getter         RepoGetter
-	EventLister    RepoEventLister
-	Lister         RepoLister
-	UserService    *RepoUserService
-	ServiceService *RepoServiceService
-	MineLister     RepoMineLister
+	AccountLister      RepoAccountLister
+	Creater            RepoCreater
```
golangcibot: `Creater` is a misspelling of `Creature` (from `misspell`)
```suggestion

```

SimonBarendse:
> Let's make sure we implement it in the CLI before releasing this. This way we can find any implementation quirks.

Good call :+1: Made a change so that we can mock the client when testing.

PR to use this in the CLI is up: https://github.com/secrethub/secrethub-cli/pull/215

PR #146, opened 2019-11-13
Add CreateAll func to dir service to create dirs recursively

SimonBarendse:
It works just like mkdir -p, creating all directories that do not
exist yet recursively and it does not error when the directories
already exist.

### DIFF
--- a/pkg/secrethub/dir.go
+++ b/pkg/secrethub/dir.go
@@ -1,15 +1,20 @@
 package secrethub
 
 import (
+	"strings"
+
 	"github.com/secrethub/secrethub-go/internals/api"
 	"github.com/secrethub/secrethub-go/internals/api/uuid"
 	"github.com/secrethub/secrethub-go/internals/errio"
+	"github.com/secrethub/secrethub-go/pkg/secretpath"
 )
 
 // DirService handles operations on directories from SecretHub.
 type DirService interface {
 	// Create a directory at a given path.
 	Create(path string) (*api.Dir, error)
+	// CreateAll creates all directories in the given path that do not exist yet.
+	CreateAll(path string) error
 	// Get returns the directory with the given ID.
 	GetByID(id uuid.UUID) (*api.Dir, error)
 	// Delete removes the directory at the given path.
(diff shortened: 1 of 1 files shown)


pkg/secrethub/dir.go
```diff
@@ -1,15 +1,20 @@
...
 	"github.com/secrethub/secrethub-go/internals/api/uuid"
 	"github.com/secrethub/secrethub-go/internals/errio"
+	"github.com/secrethub/secrethub-go/pkg/secretpath"
 )
 
 // DirService handles operations on directories from SecretHub.
 type DirService interface {
 	// Create a directory at a given path.
 	Create(path string) (*api.Dir, error)
+	// CreateAll creates all directories in the given path that do not exist yet.
```
mackenbach: Consider adding here that it doesn't error when directories already exist.

pkg/secrethub/dir.go
```diff
@@ -169,6 +174,35 @@ func (s dirService) Delete(path string) error {
...
+}
+
+func (s dirService) createAll(path string) error {
+	if len(strings.Split(path, "/")) < 3 {
+		return nil
+	}
+
+	_, err := s.GetTree(path, 0, false)
+	if err != api.ErrDirNotFound {
+		return err
```
mackenbach: When the dir already exists, doesn't this return an error? Oh wait, then it returns `nil`, correct?

Consider adding an explanatory comment here because it's not suuuper clear at a glance.

pkg/secrethub/dir.go
```diff
@@ -169,6 +174,35 @@ func (s dirService) Delete(path string) error {
...
+	}
+
+	_, err = s.Create(path)
+	if err == api.ErrDirAlreadyExists {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	return s.createAll(secretpath.Parent(path))
```
mackenbach: Could you walk me through the logic here? For instance take the path `company/repo/foo/bar/baz`, then does it call `s.Create(path) in this order:

1. `s.Create("company/repo/foo/bar/baz")`
2. `s.Create("company/repo/foo/bar")`
1. `s.Create("company/repo/foo")`

If it does call it in that order, it seems to me like it would fail, correct?

SimonBarendse: You're right! fixed it and added a test.

PR #148, opened 2019-11-21
Add credential listing

jpcoenen:


jpcoenen:
That's a good question. I first thought we just replace `List` with `Iterator`. But that does not really work in this case. We could also directly go for a stub iterator implementation (one that fetches the complete list, but serves an `Iterator` that implements `Next()`.)

SimonBarendse:
> We could also directly go for a stub iterator implementation (one that fetches the complete list, but serves an Iterator that implements Next().)

That'd be nice :+1:

jpcoenen:
Okay. Making the stub is easy. But that does not really solve the naming issue. We still have many options for that:

`List()`
`ListMine()`
`MineIterator()`
`Mine()`

@mackenbach , @florisvdg , does one of you have an opinion in this regard?

florisvdg:
If there's no other than mine, then 'mine' is implied right? So it could just be `List()`. 

Or is there a possibility in the future to list other people's credentials?

jpcoenen:
> Or is there a possibility in the future to list other people's credentials?

No, can't imagine we need that.

The main point is about what we want to do with returning an `Iterator`. For `AuditEvents` the function for getting an `Iterator` is `SecretEventIterator`, but the main reason for that is that `ListSecretEvents` is already taken. So the question is: do we want to keep it similar to `AuditEvents` (having `Iterator` in the name) or going for the more convenient `List`?

SimonBarendse:
I believe we all agreed that in the end we'd like all functions to be named `List`, correct?

Therefore, I see two possibilities now:
1. Already go to the desired end-state now, by naming the function `List` (or `ListMine`?)
2. Create an `Iterator` or `MineIterator` now, to be consistent with how other functions that return iterators are called. Add a List func, deprecate the iterator func and remove the iterator func when also doing so for the other iterator functions (e.g. for audit events).

Advantage of 1 is that there's no migration needed from the user of the function (no deprecation, no removal). Advantage of 2 is that functions are consistently named during the migration period.

Migration is as follows:

add iterator funcs => deprecate list funcs => :hourglass_flowing_sand:
=> remove list funcs => :hourglass_flowing_sand:?
=> add list funcs (which return iterators) => deprecate iterator funcs => :hourglass_flowing_sand:
=> remove iterator funcs

Where :hourglass_flowing_sand: means there's time (to migrate) between releases. Therefore, this whole migration might take a while, making it more important that functions are consistently named during the migration period.

SimonBarendse:
> Where :hourglass_flowing_sand: means there's time (to migrate) between releases.

Now that I'm writing this down I wonder whether this is really necessary?

Users can upgrade to the version that includes the deprecation, do the migration and then upgrade again to the version that includes the removal. They decide when to upgrade.
Only requirement to make this work is that we keep updating the version before removal (the one that contains the deprecation) with bugfixes.

Or do we want users to be able to get additional client functionality without having to migrate?

jpcoenen:
> add iterator funcs => deprecate list funcs => hourglass_flowing_sand
> => remove list funcs => hourglass_flowing_sand?
> => add list funcs (which return iterators) => deprecate iterator funcs => hourglass_flowing_sand
> => remove iterator funcs

Ugh. This is really tedious. Means a user has to update their code multiple times. This makes me favour just going with `List()` directly.

SimonBarendse:
> Ugh. This is really tedious. Means a user has to update their code multiple times. This makes me favour just going with List() directly.

Yes I agree. Does make it more important to start the migration for the other iterators, so that we start moving towards the consistent naming. So we might want to prioritize that a bit more.

pkg/secrethub/internals/http/client.go
```diff
@@ -141,43 +140,51 @@ func (c *Client) CreateSession(in interface{}) (*api.Session, error) {
...
 }
 
+// ListMyCredentials list all the currently authenticated account's credentials.
+func (c *Client) ListMyCredentials() ([]*api.Credential, error) {
+	var out []*api.Credential
+	rawURL := fmt.Sprintf(pathCredentials, c.base.String())
+	err := c.get(rawURL, true, &out)
+	return out, errio.Error(err)
+}
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

pkg/secrethub/credentials.go
```diff
@@ -71,6 +75,39 @@ func (s credentialService) Create(creator credentials.Creator) error {
...
+func (s credentialService) List(_ *CredentialListParams) (CredentialIterator, error) {
+	creds, err := s.client.httpClient.ListMyCredentials()
+	if err != nil {
+		return nil, err
+	}
+	return &credentialIterator{
+		credentials: creds,
+	}, nil
+}
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

pkg/secrethub/internals/http/client.go
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

pkg/secrethub/credentials.go
```diff
@@ -73,6 +73,10 @@ func (s credentialService) Create(creator credentials.Creator) error {
 	return nil
 }
 
+// CredentialListParams are the parameters than configure credential listing.
```
SimonBarendse: typo `than` => `that`

pkg/secrethub/credentials.go
```diff
@@ -71,7 +73,31 @@ func (s credentialService) Create(creator credentials.Creator) error {
...
+	currentIndex := c.currentIndex
+	if currentIndex >= len(c.credentials) {
+		return api.Credential{}, iterator.Done
+	}
+	c.currentIndex++
+	return *c.credentials[currentIndex], nil
+}
+
+// List lists all credentials of the currently authenticated account.
+func (s credentialService) List() (CredentialIterator, error) {
```
SimonBarendse: To be consistent with audit events and for simplicity for the user, creating the iterator shouldn't return an error.
We can fetch the list of credentials on the first call to `Next`.

PR #150, opened 2019-11-25
Allow setting the credential description for new credentials

jpcoenen:
For now we're not setting the credential name on a signup, as this is a backwards-incompatible change. We can later add the functionality to rename any credentials without a name.

### DIFF
--- a/internals/api/credential.go
+++ b/internals/api/credential.go
@@ -21,7 +21,7 @@ var (
 	ErrCredentialFingerprintNotUnique = errAPI.Code("fingerprint_not_unique").StatusErrorf("there are multiple credentials that start with the given fingerprint. Please use the full fingerprint", http.StatusConflict)
 	ErrInvalidVerifier                = errAPI.Code("invalid_verifier").StatusError("verifier is invalid", http.StatusBadRequest)
 	ErrInvalidCredentialType          = errAPI.Code("invalid_credential_type").StatusError("credential type is invalid", http.StatusBadRequest)
-	ErrInvalidCredentialName          = errAPI.Code("invalid_credential_name").StatusError("credential name must be between 1 and 20 characters long", http.StatusBadRequest)
+	ErrInvalidCredentialDescription   = errAPI.Code("invalid_credential_description").StatusError("credential description must be between 1 and 20 characters long", http.StatusBadRequest)
 	ErrInvalidAWSEndpoint             = errAPI.Code("invalid_aws_endpoint").StatusError("invalid AWS endpoint provided", http.StatusBadRequest)
 	ErrInvalidProof                   = errAPI.Code("invalid_proof").StatusError("invalid proof provided for credential", http.StatusUnauthorized)
 	ErrAWSAccountMismatch             = errAPI.Code("aws_account_mismatch").StatusError("the AWS Account ID in the role ARN does not match the AWS Account ID of the AWS credentials used for authentication. Make sure you are using AWS credentials that correspond to the role you are trying to add.", http.StatusUnauthorized)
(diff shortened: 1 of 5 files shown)


SimonBarendse:
From https://github.com/secrethub/secrethub-cli/pull/219 :

> On signup, this does not add a name to the credential yet because this is a breaking change. Furthermore, it degrades the experience of the signup because we ask another question. We can later add this or implement a feature add a name to a credential that does not have one yet. We should be careful with allowing renames, as this could make forensics more difficult if an account would be compromised.

1. Why would setting an extra field be a breaking change?
2. I agree that asking the user extra questions degrades the experience. What if we would auto-fill a device field with the host-name of the device in both the backup and signup scenario? Then we don't bother the user with questions in both scenarios and the credentials are better distinguishable when listing because both credentials have the device name set.
3. In this PR you suggest adding rename functionality, in the CLI PR, you warn that this might not be what we want. Could we add the device functionality to signup now so that we don't need the rename functionality?

florisvdg:
> For now we're not setting the credential name on a signup
It's probably fine if it's never set. The user doesn't care about that until there are multiple credentials. So either have an `is_default` flag that allows the `name` to be empty, or hardcode something like `default` as the `name`.

jpcoenen:
> Why would setting an extra field be a breaking change?

Because it changes the signature for `Signup()`.

> In this PR you suggest adding rename functionality, in the CLI PR, you warn that this might not be what we want. Could we add the device functionality to signup now so that we don't need the rename functionality?

The warning concerns allowing it without restriction. Going from no name to a name should not be a big problem.

> It's probably fine if it's never set. The user doesn't care about that until there are multiple credentials. So either have an is_default flag that allows the name to be empty, or hardcode something like default as the name.

For `secrethub init` with a backup code, I have implement a default credential name: the hostname. It is suggested as:
```
What is the name of this device? [MY-HOSTNAME]
```
In that way, the cognitive overhead is as small as possible.

jpcoenen:
> Did we land on having a credential name? We also discussed having an optional description or a separate device (name) field.

You're right. I'll change it to `Description` tomorrow.

SimonBarendse:
>>In this PR you suggest adding rename functionality, in the CLI PR, you warn that this might not be what we want. Could we add the device functionality to signup now so that we don't need the rename functionality?
>
> The warning concerns allowing it without restriction. Going from no name to a name should not be a big problem.

Woops, you're right. Sorry about that

> >Why would setting an extra field be a breaking change?
>
> Because it changes the signature for Signup(). This is a big downside of having client functions that accept multiple argument instead of a single struct with all the parameters.

:+1:
I think we should consider adding a device field. As this can be automatically detected (as you suggest with hostname) we can detect and set this in the client, so that we don't require an extra argument for `Signup` (and credential create) func. It's not possible for the user to override the device name then though (as you suggest with the prompt that has a default filled in).

SimonBarendse:
Discussed the name/description/device options in person.

As the hostname isn't always descriptive enough, we want to prompt the user for input (with suggestion). Therefore, we think name or description fits better than device.

As we don't want to require a name/description and it doesn't have to be unique, we think description fits better than name.

internals/api/credential.go
```diff
@@ -21,7 +21,7 @@ var (
 	ErrCredentialFingerprintNotUnique = errAPI.Code("fingerprint_not_unique").StatusErrorf("there are multiple credentials that start with the given fingerprint. Please use the full fingerprint", http.StatusConflict)
 	ErrInvalidVerifier                = errAPI.Code("invalid_verifier").StatusError("verifier is invalid", http.StatusBadRequest)
 	ErrInvalidCredentialType          = errAPI.Code("invalid_credential_type").StatusError("credential type is invalid", http.StatusBadRequest)
-	ErrInvalidCredentialName          = errAPI.Code("invalid_credential_name").StatusError("credential name must be between 1 and 20 characters long", http.StatusBadRequest)
+	ErrInvalidCredentialDescription   = errAPI.Code("invalid_credential_description").StatusError("credential description must be between 1 and 20 characters long", http.StatusBadRequest)
```
SimonBarendse: The validation func checks for length >32 chars, but this error message still says 20 chars

SimonBarendse: not introduced in this PR though

PR #154, opened 2019-11-27
Release v0.24.0

SimonBarendse:
### Added
- Add shorthand for generating random alphanumeric characters (#142)
- Add `CreateAll` to create dirs recursively (#146)
- Add `Exists` for directories (#149, #153)
- Add functionality to iteratively fetch audit events (#144)
- Validate custom server URL when constructing the client (fail fast) (#144)


pkg/secrethub/repo.go
```diff
@@ -14,6 +15,21 @@ type RepoService interface {
...
+	//  	event, err := iter.Next()
+	//  	if err == iterator.Done {
+	//  		break
+	//  	} else if err != nil {
+	//  		// Handle error
+	//  	}
+	//
+	//  	// Use event
+	//  }
+	EventIterator(path string, _ *AuditEventIteratorParams) AuditEventIterator
```
mackenbach: Why do we have an unnamed parameter here? Would `params` be more descriptive?

pkg/secrethub/repo.go
```diff
@@ -107,6 +123,38 @@ func (s repoService) ListEvents(path string, subjectTypes api.AuditSubjectTypeLi
 	return events, nil
 }
 
+// EventIterator returns an iterator that retrieves all audit events for a given repo.
+//
+// Usage:
+//  iter, err := client.Repos().EventIterator(path, &secrethub.AuditEventIteratorParams{})
```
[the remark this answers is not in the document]
SimonBarendse: You're right. I've opened https://github.com/secrethub/secrethub-go/pull/158 to fix this.

pkg/secrethub/repo.go
SimonBarendse: Yes, `params` as a name could also work.

The paramater is unnamed because it is not currently being used; there are no parameters to pass to the event iterator yet. The paramater is already added so that we can add parameters to the params struct without requiring a breaking change.

PR #161, opened 2019-12-11
Verify client version is set correctly on release branches

SimonBarendse:
To prevent [this](https://github.com/secrethub/secrethub-go/blob/v0.24.0/pkg/secrethub/client_version.go#L5)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  build:
+  test:
     docker:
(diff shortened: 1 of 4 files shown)


florisvdg:
We could also make a GitHub Action that actually corrects it.

SimonBarendse:
> Cool. What about hotfix branches?

> We could also make a GitHub Action that actually corrects it.

Yes, both interesting ideas! :+1:
Shall we include this first and then see where to take it from there?

PR #162, opened 2019-12-13
Reordered interface functions to CRUDL

Marton6:


pkg/secrethub/secret.go
```diff
@@ -8,12 +8,31 @@ import (
...
+	// ReadString is a convenience function to get the secret data as a string.
+	//
+	// See .Versions() for more elaborate use.
+	ReadString(path string) (string, error)
 	// Exists returns whether a secret exists on the given path.
 	Exists(path string) (bool, error)
 	// Get retrieves a Secret.
 	Get(path string) (*api.Secret, error)
+	// Versions returns a SecretVersionService.
+	Versions() SecretVersionService
```
mackenbach: What made you put `Versions` here? In other places we have the 'sub services' at the end I noticed.

Marton6: I thought it belonged there since it has to do mostly with reading secrets, however it  would indeed be more consistent to have it at the end.

PR #163, opened 2019-12-20
Add iterator functions

Marton6:


pkg/secrethub/acl.go
```diff
@@ -286,3 +292,93 @@ func (c *Client) getAccessLevel(path api.BlindNamePath, accountName api.AccountN
...
+	return &accessRuleIterator{
+		index: 0,
+		data:  data,
+		err:   err,
+	}
+}
+
+// AccessRuleIteratorParams specify parameters used when listing access rules.
+type AccessRuleIteratorParams struct {
+	depth     int  // depth defines the depth of traversal for the iterator, -1 means listing all subdirectories.
```
SimonBarendse: `int` has a `zero-value` of 0. As the zero value will be used when constructing the params struct without specifying the field, the zero-value will act as the default value.

For this function, I think returning access rules for all subdirectories (currently specified with `-1`) would be a better default for this function.

To be able to use a value other than zero as the default value, we could change the type to `*int` and use the default value when the value isn't explicitly set (when the value is the zero-value of a pointer value: `nil`).

If we want all subdirectories to be the default, we can even change the type to `*uint`, so that negative values are not allowed.

pkg/secrethub/org.go
```diff
@@ -17,6 +18,8 @@ type OrgService interface {
 	Delete(name string) error
 	// ListMine returns the organizations of the current user.
 	ListMine() ([]*api.Org, error)
+	// Iterator returns an iterator that lists all organizations of the current users.
```
SimonBarendse: Typo? There is only one current user.

pkg/secrethub/org.go
```diff
@@ -73,3 +76,41 @@ func (s orgService) Members() OrgMemberService {
 func (s orgService) ListMine() ([]*api.Org, error) {
 	return s.client.httpClient.ListMyOrgs()
 }
+
+// Iterator returns an iterator that lists all organizations of the current users.
```
SimonBarendse: Typo? There is only one current user.

pkg/secrethub/org_member.go
```diff
@@ -112,3 +115,41 @@ func (s orgMemberService) Update(org string, username string, role string) (*api
...
+func (s orgMemberService) Iterator(org string, _ *OrgMemberIteratorParams) OrgMemberIterator {
+	data, err := s.List(org)
+	return &orgMemberIterator{
+		index: 0,
+		data:  data,
+		err:   err,
+	}
+}
+
+// OrgMemberIteratorParams defines parameters used when listing OrgMembers.
```
SimonBarendse: When writing godoc, we usually try to use natural language as much as possible, rather than naming the specific type returned. In this case, that would mean saying `when listing members of the organization` rather than `when listing OrgMembers`

pkg/secrethub/repo.go
```diff
@@ -17,8 +18,12 @@ type RepoService interface {
 	Delete(path string) error
 	// List retrieves all repositories in the given namespace.
 	List(namespace string) ([]*api.Repo, error)
+	// Iterator returns a new Iterator that retrieves all repos in the given namespace.
```
SimonBarendse: `Iterator` => `iterator`?

pkg/secrethub/org.go
```diff
@@ -73,3 +76,41 @@ func (s orgService) Members() OrgMemberService {
...
+
+// RepoIteratorParams defines parameters used when listing repos.
+type OrgIteratorParams struct{}
+
+// RepoIterator iterates over repositories.
+type OrgIterator interface {
+	Next() (api.Org, error)
+}
+
+type orgIterator struct {
```
mackenbach: Nice work Márton! Just had a quick read through and one question popped in my mind:

Are these iterators implemented in a thread-safe / goroutine safe way? And more importantly, should they?

If you have questions on this, discuss it with @SimonBarendse I know he has thoughts about this 😄

pkg/secrethub/acl.go
Marton6: Good point! I'll use `*uint`.

pkg/secrethub/org_member.go
Marton6: Thanks for the tip! I'll rephrase the comment.

pkg/secrethub/org.go
Marton6: Good question(s)! I will discuss this with @SimonBarendse on Monday.

pkg/secrethub/acl.go
```diff
@@ -316,8 +319,8 @@ func (s accessRuleService) Iterator(path string, params *AccessRuleIteratorParam
 
 // AccessRuleIteratorParams specify parameters used when listing access rules.
 type AccessRuleIteratorParams struct {
-	depth     int  // depth defines the depth of traversal for the iterator, -1 means listing all subdirectories.
-	ancestors bool // ancestors defines whether the iterator should also list access rules of parent directories.
+	depth     *uint  	// depth defines the depth of traversal for the iterator, nil means listing all subdirectories.
```
SimonBarendse: To be able to set a value on these fields, they should be exported.

pkg/secrethub/org.go
SimonBarendse: Yes, would be nice to make it safe for concurrent access. Especially since this is also done for the audit event iterator, so to be consistent it would be good to also make these iterators safe for concurrent access.

Have a look here: https://github.com/secrethub/secrethub-go/blob/v0.25.0/pkg/secrethub/iterator/iterator.go#L40 how this is done for the audit iterator.

Another possibility is to already adopt the iterator/paginator pattern and use [`iterator.Iterator`](https://github.com/secrethub/secrethub-go/blob/v0.25.0/pkg/secrethub/iterator/iterator.go#L22), which already provides the safety for concurrent access. The paginator could return a single page for now.
This will also make it easier to switch to paginated pages later, as we'll only have to swap out the paginator.

pkg/secrethub/iterator/paginator.go
```diff
@@ -0,0 +1,31 @@
+package iterator
+
+type paginator struct {
+	fetched bool
+	fetch   func() ([]interface{}, error)
+}
+
+// PaginatorConstructorWithFetch returns a new paginator constructor that constructs a paginator
+// with the provided fetch function.
+func PaginatorConstructorWithFetch(fetch func() ([]interface{}, error)) PaginatorConstructor {
```
SimonBarendse: Maybe call this `NewPaginatorFactory`? Or `NewSinglePagePaginatorFactory` (as all paginators fetch pages, the unique thing about this paginator is that is fetches everything in one page.

mackenbach: Guys, just saw this pass by and it looks like you're creating a ConstructorConstructor to inject fetch functionality. This seems pretty convoluted and looks like it's for testing purposes. 

If that's the case, think carefully about exposing a public function in this package only for testing purposes.

pkg/secrethub/acl.go
Marton6: Good point! :+1:

pkg/secrethub/iterator/paginator.go
```diff
@@ -0,0 +1,31 @@
...
+	fetched bool
+	fetch   func() ([]interface{}, error)
+}
+
+// PaginatorFactory returns a paginator constructor that constructs a paginator
+// with the provided fetch function.
+func PaginatorFactory(fetch func() ([]interface{}, error)) PaginatorConstructor {
+	return func() (Paginator, error) {
+		return &paginator{
+			fetched:false,
```
golangcibot: File is not `goimports`-ed (from `goimports`)

pkg/secrethub/iterator/iterator_test.go
```diff
@@ -0,0 +1,27 @@
+package iterator
+
+import (
+	"github.com/secrethub/secrethub-go/internals/assert"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

pkg/secrethub/audit_test.go
```diff
@@ -16,7 +16,7 @@ type fakeAuditPaginator struct {
 
 func (pag *fakeAuditPaginator) Next() ([]interface{}, error) {
 	if pag.returned {
-		return []interface{}{}, nil
+		return make([]interface{}, 0), nil
```
SimonBarendse: Can you explain why you changed this?

pkg/secrethub/repo.go
```diff
@@ -309,3 +316,113 @@ func (c *Client) getRepoIndexKey(repoPath api.RepoPath) (*crypto.SymmetricKey, e
...
+					}
+					return res, nil
+				},
+			),
+		),
+	}
+}
+
+// AccountIterator returns a new iterator that retrieves all accounts in the given namespace.
+func (s repoService) AccountIterator(namespace string, params *RepoIteratorParams) AccountIterator {
```
SimonBarendse: `ListAccounts` has a path as argument, while this iterator has a namespace as argument. How does that work?

pkg/secrethub/repo.go
```diff
@@ -309,3 +316,113 @@ func (c *Client) getRepoIndexKey(repoPath api.RepoPath) (*crypto.SymmetricKey, e
 
 	return repoIndexKey, nil
 }
+
+// Iterator returns a new iterator that retrieves all repos in the given namespace.
+func (s repoService) Iterator(namespace string, params *RepoIteratorParams) RepoIterator {
```
SimonBarendse: We also discussed having iteration of repos in a given namespace and iteration of all repos you have access to done in one iterator. Why did you choose for separating them?

pkg/secrethub/acl.go
```diff
@@ -363,31 +385,15 @@ type AccessLevelIterator interface {
...
 
 type accessLevelIterator struct {
-	index int
-	data  []*api.AccessLevel
-	err   error
+	iterator iterator.Iterator
 }
 
-// Next returns the next access level or iterator.Done if the all of them have been returned.
+// Next returns the next access rule or iterator.Done if the all of them have been returned.
```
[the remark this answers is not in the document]
Marton6: Yes, that was a typo.

pkg/secrethub/audit_test.go
Marton6: I probably did it by mistake.

pkg/secrethub/acl.go
```diff
@@ -286,3 +292,108 @@ func (c *Client) getAccessLevel(path api.BlindNamePath, accountName api.AccountN
...
+	if params == nil {
+		depth = -1
+		ancestors = false
+	} else if params.Depth == nil {
+		depth = -1
+		ancestors = params.Ancestors
+	} else {
+		depth = int(*params.Depth)
+		ancestors = params.Ancestors
+	}
```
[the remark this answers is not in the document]
Marton6: Great idea, thanks!

pkg/secrethub/repo.go
Marton6: That is another typo.

pkg/secrethub/repo.go
Marton6: Because if namespace would be an optional param in `RepoIteratorParams` then this set of parameters would not make sense in the context of the `MeService`'s `RepoIterator` method. I think it is better if the `RepoIteratorParams` are the params used whenever iterating over repos regardless of service.

SimonBarendse: > Because if namespace would be an optional param in RepoIteratorParams then this set of parameters would not make sense in the context of the MeService's RepoIterator method.

Why not? The iterator in the me service is returning all repositories the authenticated user has access to. The namespace parameter could limit the response to those repositories that are in the given namespace.
You can see it as a filter.

pkg/secrethub/acl.go
```diff
@@ -296,18 +296,17 @@ func (c *Client) getAccessLevel(path api.BlindNamePath, accountName api.AccountN
...
-	} else {
+		params = &AccessRuleIteratorParams{}
+	}
+
+	var depth int
+	if params.Depth != nil {
 		depth = int(*params.Depth)
-		ancestors = params.Ancestors
+	} else {
+		depth = -1
```
SimonBarendse: We can even go one step further and do:

```Go
depth := -1
if params.Depth != nil {
  depth = int(*params.Depth)
}
```

Which do you prefer?

pkg/secrethub/repo.go
Marton6: Okay that makes sense, apparently the functionality of iterator in the me service wasn't clear to me.

pkg/secrethub/repo.go
```diff
@@ -309,3 +314,112 @@ func (c *Client) getRepoIndexKey(repoPath api.RepoPath) (*crypto.SymmetricKey, e
...
+					}
+					return res, nil
+				},
+			),
+		),
+	}
+}
+
+// RepoIteratorParams defines parameters used when listing repos.
+type RepoIteratorParams struct{
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
type RepoIteratorParams struct {
```

pkg/secrethub/repo.go
```diff
@@ -309,3 +314,112 @@ func (c *Client) getRepoIndexKey(repoPath api.RepoPath) (*crypto.SymmetricKey, e
...
+	}
+
+	return &repoIterator{
+		iterator: iterator.New(
+			iterator.PaginatorFactory(
+				func() ([]interface{}, error) {
+					var err error
+					var repos []*api.Repo
+					if params.Namespace == nil {
+						repos, err = s.client.httpClient.ListMyRepos()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/secrethub/acl.go
Marton6: I prefer the shortest version

pkg/secrethub/secret_version.go
```diff
@@ -272,3 +276,58 @@ func (c *Client) decryptSecretVersions(encVersions ...*api.EncryptedSecretVersio
 
 	return versions, nil
 }
+
+// Iterator returns a new iterator that retrieves all accounts in the given namespace.
```
SimonBarendse: This comment still says accounts.

SimonBarendse: Maybe add to the comment that `IncludeSensitiveData` should be set if you want to retrieve the secret data, to make this extra clear and prevent confusion when the sensitive data isn't set.

SimonBarendse: and to the interface as well.

pkg/secrethub/service.go
```diff
@@ -115,3 +118,52 @@ func (s serviceService) List(path string) ([]*api.Service, error) {
 	repoServiceService := newRepoServiceService(s.client)
 	return repoServiceService.List(path)
 }
+
+// Iterator returns an iterator that lists all service accounts in a given repository.
```
SimonBarendse: Another case where the comment is still there from a copy-paste. Could you check all iterators to see if there are more cases?

pkg/secrethub/secret_version.go
Marton6: Good point!

SimonBarendse:
@Marton6 We still need to deprecate the list functions: https://github.com/secrethub/secrethub-go/pull/163#pullrequestreview-335231759

pkg/secrethub/secret.go
```diff
@@ -47,6 +47,7 @@ type SecretService interface {
 	//  }
 	EventIterator(path string, _ *AuditEventIteratorParams) AuditEventIterator
 	// ListEvents retrieves all audit events for a given secret.
+	// Deprecated: Use iterator function instead.
```
SimonBarendse: This one we cannot deprecate yet, as the audit subject type filter is not supported yet on the iterator.

Marton6: Fixed!

SimonBarendse:
@mackenbach Is anything holding this from being merged?

mackenbach:
Nope, I just left the timing of the merge to @Marton6 because it looked like there was a dependency. Merged it now.

PR #164, opened 2019-12-20
Add examples to godoc

Marton6:


pkg/randchar/example_test.go
```diff
@@ -0,0 +1,42 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	print(string(val))
+}
+
+// Generate a 10 character alphanumeric string containing lowercase letters and digits.
+func ExampleRand_Generate_customCharset() {
+	customCharset := NewCharset("abcdefghijklmnopqrstuvwxyz0123456789")
```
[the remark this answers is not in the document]
SimonBarendse: We could have for example this `Lowercase.Add(Numeric)` and an example of a hexadecimal charset: `NewCharset("0123456789ABCDEF")`

pkg/randchar/example_test.go
```diff
@@ -0,0 +1,42 @@
...
+		log.Fatal(err)
+	}
+
+	val, err := rand.Generate(15)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	print(string(val))
+}
```
SimonBarendse: Do we also want to add an example of the most simple usecase: generating a random slice of alphanumeric characters:
```Go
val, err := randchar.Generate(30)
if err != nil {
  log.Fatal(err)
}
print(string(val))
```

pkg/secrethub/example_test.go
```diff
@@ -11,6 +11,26 @@ import (
...
+func ExampleNewClient() {
+	client, err := NewClient()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Printf("Client with ID: %s", client.account.AccountID.String())
+}
+
+// Create a new aws Client.
```
SimonBarendse: Maybe extend this explanation a bit more.

e.g.
```
Create a new client that uses native AWS services to handle encryption and authentication.
```

pkg/randchar/example_test.go
```diff
@@ -0,0 +1,42 @@
+package randchar
```
SimonBarendse: Shall we change the package to the external test package `randchar_test`?

I think this will make the examples more representative for how the `randchar` package is consumed: from outside the package itself. The examples will then also import the `randchar` package and use the prefix (e.g. `randchar.Min` instead of just `Min`). Furthermore, we ensure the examples don't use unexported code.

SimonBarendse:
PS: CI is renamed from `workflow` to `pipeline` on `develop`. Merging `origin/develop` to your branch should fix the required `pipeline` test to pass.

pkg/randchar/example_test.go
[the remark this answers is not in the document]
Marton6: Let's show both. I had no better idea for a custom charset to demonstrate this functionality with, but hexadecimal seems like a good example.

pkg/randchar/example_test.go
Marton6: Good idea! We could also do this with the Client examples.

pkg/secrethub/example_test.go
```diff
@@ -0,0 +1,121 @@
...
+	client, err := secrethub.NewClient()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// use the client
+	_, err = client.Repos().Create("workspace/repo")
+	if err != nil {
+		log.Fatal(err)
+	}
```
Marton6: Do you have a better suggestion of how to use the client in the example in which we create it?

SimonBarendse: Maybe a `client.Secrets().ReadString`?

PR #167, opened 2019-12-26
Use api.IsErrNotFound instead of checking for errors directly

Marton6:


pkg/secrethub/secret.go
```diff
@@ -171,7 +171,7 @@ func (s secretService) Write(path string, data []byte) (*api.SecretVersion, erro
 	}
 
 	key, err := s.client.getSecretKey(secretPath)
-	if api.IsErrNotFound(err) {
+	if err == api.ErrSecretNotFound {
 		return s.client.createSecret(secretPath, data)
 	} else if err == api.ErrNoOKSecretKey {
 		key, err = s.client.createSecretKey(secretPath)
```
Marton6: Can and should the direct error check be avoided in this case?

pkg/secrethub/account_key.go
```diff
@@ -49,7 +49,7 @@ func (s accountKeyService) Create(verifier credentials.Verifier, encrypter crede
 // Exists returns whether an account key exists for the client's credential.
 func (s accountKeyService) Exists() (bool, error) {
 	_, err := s.client.getAccountKey()
-	if err == api.ErrAccountKeyNotFound || err == api.ErrCredentialNotKeyed {
```
SimonBarendse: I believe `api.ErrCredentialNotKeyed` is a 409 (conflict), not a not found error.

pkg/secrethub/secret.go
```diff
@@ -171,7 +171,7 @@ func (s secretService) Write(path string, data []byte) (*api.SecretVersion, erro
 	}
 
 	key, err := s.client.getSecretKey(secretPath)
-	if err == api.ErrSecretNotFound {
+	if api.IsErrNotFound(err) {
```
SimonBarendse: `api.ErrNoOKSecretKey` appears to be a not found error as well. I believe we should check for a specific error in this case.

pkg/secrethub/secret.go
SimonBarendse: In this case, it's part of the functionality to check more specifically than is something not found; namely, we distinguish between two cases:
- the secret does not exist yet
- the secret exists, but the secret key has to be refreshed.

pkg/secrethub/account_key.go
mackenbach: When does `api.ErrCredentialNotKeyed` occur? I'm hoping only on a POST to e.g. credentials?

SimonBarendse: > I believe `api.ErrCredentialNotKeyed` is a 409 (conflict), not a not found error.

This is incorrect; `api.ErrCredentialNotKeyed` is a not found error.

It is returned when the account is keyed (has a public account key) but there is no private key stored that is encrypted for the credential with which the request is authenticated.

So, the proposed change provides the desired behavior :+1:

PR #168, opened 2019-12-28
Add mapping function for charsets

Marton6:


pkg/randchar/generator.go
```diff
@@ -244,6 +244,30 @@ func NewCharset(characters string) Charset {
 	}
 }
 
+// GetCharset returns the charset with the specified name.
+func GetCharset(charsetName string) Charset {
```
SimonBarendse: Maybe call this `CharsetByName`, to distinguish from e.g. creating a charset from all characters in the string (as `NewCharset` does).

pkg/randchar/generator.go
```diff
@@ -244,6 +244,30 @@ func NewCharset(characters string) Charset {
...
+	case "symbols":
+		return Symbols
+	case "alphanumeric":
+		return Alphanumeric
+	case "all":
+		return All
+	case "similar":
+		return Similar
+	default:
+		return Alphanumeric
```
SimonBarendse: Do we want to use a default? This means we return `Alphanumeric` for every charset name that we do not know (yet). For example: `"numbers"`.

Other examples would include making a typo, e.g. `"symblos"`

This could result in hard-to-debug (or even undiscovered) bugs.

Do we want to return an error instead?

pkg/randchar/generator.go
```diff
@@ -244,6 +244,30 @@ func NewCharset(characters string) Charset {
 	}
 }
 
+// GetCharset returns the charset with the specified name.
+func GetCharset(charsetName string) Charset {
+	switch charsetName {
+	case "numeric":
```
SimonBarendse: Could you add aliases `numbers` and `digits`?

pkg/randchar/generator.go
mackenbach: returning `Charset, bool` would solve that I guess? No need for an error I'd say.

Marton6: Good point! I'll return `Charset, bool`.

pkg/randchar/generator.go
```diff
@@ -244,6 +244,30 @@ func NewCharset(characters string) Charset {
 	}
 }
 
+// GetCharset returns the charset with the specified name.
```
SimonBarendse: godoc is still named `GetCharset`, while the function has been renamed to `CharsetByName`.

Marton6: Fixed it!

PR #172, opened 2020-02-25
Update secret not found error message

Marton6:
The `secret not found` error will now include the path of the secret. For example:
```
Encountered an error: cannot find secret: test/test/test, encountered error: Repo not found (server.repo_not_found)  (client.secret_not_found)
```

### DIFF
--- a/pkg/secrethub/secret_version.go
+++ b/pkg/secrethub/secret_version.go
@@ -2,6 +2,7 @@ package secrethub
 
 import (
 	"fmt"
+	"net/http"
 
 	"github.com/secrethub/secrethub-go/pkg/secrethub/iterator"
 
@@ -21,7 +22,7 @@ var (
 	ErrSecretTooBig         = errClient.Code("secret_too_big").Error(fmt.Sprintf("maximum size of a secret is %s", units.BytesSize(MaxSecretSize)))
 	ErrEmptySecret          = errClient.Code("empty_secret").Error("secret is empty")
 	ErrCannotWriteToVersion = errClient.Code("cannot_write_version").Error("cannot (over)write a specific secret version, they are append only")
-	ErrSecretNotFound       = errClient.Code("secret_not_found").ErrorPref("cannot find secret: %s, error: %v")
+	ErrSecretNotFound       = errClient.Code("secret_not_found").StatusErrorPref("cannot find secret: %s, error: %v", http.StatusNotFound)
 )
 
 // SecretVersionService handles operations on secret versions from SecretHub.
--- a/pkg/secrethub/secret_version.go
+++ b/pkg/secrethub/secret_version.go
@@ -24,7 +24,7 @@ var (
 
 type errSecretNotFound struct {
 	path api.SecretPath
-	err error
+	err  error
 }
 
 func (e *errSecretNotFound) Error() string {
--- a/pkg/secrethub/secret_version.go
+++ b/pkg/secrethub/secret_version.go
@@ -110,7 +110,7 @@ func (s secretVersionService) get(path api.SecretPath, withData bool) (*api.Secr
 
 	encVersion, err := s.client.httpClient.GetSecretVersion(blindName, versionParam, withData)
 	if api.IsErrNotFound(err) {
-		return nil, &errSecretNotFound{path:path, err:err}
+		return nil, &errSecretNotFound{path: path, err: err}
 	} else if err != nil {
 		return nil, errio.Error(err)
 	}
--- a/pkg/secrethub/secret_version.go
+++ b/pkg/secrethub/secret_version.go
@@ -93,7 +93,7 @@ func (s secretVersionService) Delete(path string) error {
 func (s secretVersionService) get(path api.SecretPath, withData bool) (*api.SecretVersion, error) {
 	blindName, err := s.client.convertPathToBlindName(path)
 	if api.IsErrNotFound(err) {
-		return nil, &errSecretNotFound{path:path, err:err}
+		return nil, &errSecretNotFound{path: path, err: err}
 	} else if err != nil {
 		return nil, errio.Error(err)
 	}
--- a/pkg/secrethub/secret_version.go
+++ b/pkg/secrethub/secret_version.go
@@ -20,8 +20,20 @@ var (
 	ErrSecretTooBig         = errClient.Code("secret_too_big").Error(fmt.Sprintf("maximum size of a secret is %s", units.BytesSize(MaxSecretSize)))
 	ErrEmptySecret          = errClient.Code("empty_secret").Error("secret is empty")
 	ErrCannotWriteToVersion = errClient.Code("cannot_write_version").Error("cannot (over)write a specific secret version, they are append only")
-	ErrSecretNotFound       = errClient.Code("secret_not_found").ErrorPref("cannot find secret: \"%s\": %v")
 )
+
+type errSecretNotFound struct {
+	path api.SecretPath
+	err error
+}
+
+func (e *errSecretNotFound) Error() string {
+	return fmt.Sprintf("cannot find secret: \"%s\": %v", e.path, e.err)
+}
+
+func (e *errSecretNotFound) Unwrap() error {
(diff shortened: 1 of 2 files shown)


pkg/secrethub/secret_version.go
```diff
@@ -21,6 +21,7 @@ var (
 	ErrSecretTooBig         = errClient.Code("secret_too_big").Error(fmt.Sprintf("maximum size of a secret is %s", units.BytesSize(MaxSecretSize)))
 	ErrEmptySecret          = errClient.Code("empty_secret").Error("secret is empty")
 	ErrCannotWriteToVersion = errClient.Code("cannot_write_version").Error("cannot (over)write a specific secret version, they are append only")
+	ErrSecretNotFound       = errClient.Code("secret_not_found").ErrorPref("cannot find secret: %s, error: %v")
```
mackenbach: If it's an `ErrorPref`, should it be public?

Marton6: I think it should, just to be consistent with the other errors in the client.

pkg/secrethub/secret_version.go
```diff
@@ -115,7 +117,13 @@ func (s secretVersionService) GetWithData(path string) (*api.SecretVersion, erro
 		return nil, errio.Error(err)
 	}
 
-	return s.get(secretPath, true)
+	secretVersion, err := s.get(secretPath, true)
+	if api.IsErrNotFound(err) {
```
SimonBarendse: How does this impact the use of `api.IsErrNotFound`? Will the wrapped error that we now return change the output of that function? Where is it being used? Is that a problem?

pkg/secrethub/secret_version.go
```diff
@@ -21,6 +22,7 @@ var (
 	ErrSecretTooBig         = errClient.Code("secret_too_big").Error(fmt.Sprintf("maximum size of a secret is %s", units.BytesSize(MaxSecretSize)))
 	ErrEmptySecret          = errClient.Code("empty_secret").Error("secret is empty")
 	ErrCannotWriteToVersion = errClient.Code("cannot_write_version").Error("cannot (over)write a specific secret version, they are append only")
+	ErrSecretNotFound       = errClient.Code("secret_not_found").StatusErrorPref("cannot find secret: %s, error: %v", http.StatusNotFound)
```
SimonBarendse: Usually we concatenate errors without saying `error: ` again, so that we don't repeat ourselves. The error can simply be: `cannot find secret "path/to/secret": problem why the secret cannot be found`

pkg/secrethub/secret_version.go
```diff
@@ -115,7 +117,13 @@ func (s secretVersionService) GetWithData(path string) (*api.SecretVersion, erro
 		return nil, errio.Error(err)
 	}
 
-	return s.get(secretPath, true)
+	secretVersion, err := s.get(secretPath, true)
+	if api.IsErrNotFound(err) {
+		return nil, ErrSecretNotFound(path, err)
```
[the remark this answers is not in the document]
Marton6: We discussed this in person.
(Yes, this error should also be returned when not fetching data.
List already errors with the path of the secret included.)

pkg/secrethub/secret_version.go
Marton6: Opened a PR on the CLI to use `api.IsErrNotFound`:
https://github.com/secrethub/secrethub-cli/pull/252 
It should be merged first.

The current `api.IsErrNotFound` does not work for this new error. However I am not sure how to elegantly check if an error has been created with `ErrSecretNotFound`.

SimonBarendse: If you make `ErrSecretNotFound` a struct you can use [`errors.As`](https://golang.org/pkg/errors/#As). See https://blog.golang.org/go1.13-errors for an example.

pkg/secrethub/secret_version.go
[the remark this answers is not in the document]
SimonBarendse: I thought we concluded we wanted to make `list` return the same error as `get` for consistency.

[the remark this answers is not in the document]
SimonBarendse: What do you think?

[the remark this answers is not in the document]
Marton6: That's right, I forgot to mention it, but I think it should be done in a separate PR.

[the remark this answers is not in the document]
SimonBarendse: Why? They're super related changes, right?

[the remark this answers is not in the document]
SimonBarendse: I would want to release them both in the same release, therefore also merge them in the same PR.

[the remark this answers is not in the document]
Marton6: Yes, they are related changes and it would be nice to have them both, but I think this feature has a higher priority than modifying the error returned by the list command for consistency, and the latter requires a couple more design decisions that we might need to discuss.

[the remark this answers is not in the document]
SimonBarendse: Oke, if you think it will require a lot extra time, we can postpone it a bit for now. Can you create a ticket to make sure this will eventually be picked up?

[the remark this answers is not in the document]
Marton6: Yes, good point :+1:

pkg/secrethub/secret_version.go
Marton6: Wouldn't that be inconsistent with regards to how we currently create and handle errors?
I think that for now we could (inelegantly) check the namespace and code of the error and add error wrapping to the `errio` package later, to simplify this check. What do you think, @SimonBarendse ?

SimonBarendse: @Marton6 You're right that this is inconsistent with our current error checking. To get dynamic error messages however, we do have to deviate from our current error checking method. Your proposal to check for namespace and code also deviates from our current approach.

What might be important to know is that we want to get rid of the `errio` package (see also https://github.com/secrethub/secrethub-go/pull/172#discussion_r384629641), so using something different here can be the start of moving away from the `errio` package.

pkg/secrethub/secret_version.go
```diff
@@ -23,6 +22,19 @@ var (
 	ErrCannotWriteToVersion = errClient.Code("cannot_write_version").Error("cannot (over)write a specific secret version, they are append only")
 )
 
+type errSecretNotFound struct {
+	path api.SecretPath
+	err error
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	err  error
```

pkg/secrethub/secret_version.go
```diff
@@ -80,7 +92,9 @@ func (s secretVersionService) Delete(path string) error {
 // get gets a version of a secret. withData specifies whether the encrypted data should be retrieved.
 func (s secretVersionService) get(path api.SecretPath, withData bool) (*api.SecretVersion, error) {
 	blindName, err := s.client.convertPathToBlindName(path)
-	if err != nil {
+	if api.IsErrNotFound(err) {
+		return nil, &errSecretNotFound{path:path, err:err}
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		return nil, &errSecretNotFound{path: path, err: err}
```

pkg/secrethub/secret_version.go
```diff
@@ -95,7 +109,9 @@ func (s secretVersionService) get(path api.SecretPath, withData bool) (*api.Secr
 	}
 
 	encVersion, err := s.client.httpClient.GetSecretVersion(blindName, versionParam, withData)
-	if err != nil {
+	if api.IsErrNotFound(err) {
+		return nil, &errSecretNotFound{path:path, err:err}
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
		return nil, &errSecretNotFound{path: path, err: err}
```

Marton6:
This PR requires the go version of the client to be updated to 1.13

PR #174, opened 2020-03-03
Add human-readable charset

Marton6:
Added the `human-readable` charset defined as `alphanumeric`-`similar`.


pkg/randchar/generator.go
```diff
@@ -263,6 +263,8 @@ func CharsetByName(charsetName string) (Charset, bool) {
 		return All, true
 	case "similar":
 		return Similar, true
+	case "human-readable":
+		return Alphanumeric.Subtract(Similar), true
```
SimonBarendse: For consistency, shall we create a global variable for the human-readable charset as well?

PR #175, opened 2020-03-05
Use + instead of fmt.Sprintf for simple string concatenation

Marton6:
Replaced `fmt.Sprintf` with `+` where it does not affect readability.
See also: https://github.com/secrethub/secrethub-go/pull/58#discussion_r263422145

### DIFF
--- a/internals/api/paths.go
+++ b/internals/api/paths.go
@@ -2,7 +2,7 @@ package api
 
 import (
 	"encoding/base64"
-	"fmt"
+	"strconv"
 	"strings"
 
 	"bitbucket.org/zombiezen/cardcpx/natsort"
@@ -155,7 +155,7 @@ func (sp SecretPath) AddVersion(version int) (SecretPath, error) {
 		return SecretPath(""), ErrPathAlreadyHasVersion
 	}
 
-	p := SecretPath(fmt.Sprintf("%s:%d", sp.String(), version))
+	p := SecretPath(sp.String() +":"+ strconv.Itoa(version))
 	err := p.Validate()
 	if err != nil {
 		return SecretPath(""), err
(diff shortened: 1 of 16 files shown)


internals/api/paths.go
```diff
@@ -155,7 +155,7 @@ func (sp SecretPath) AddVersion(version int) (SecretPath, error) {
 		return SecretPath(""), ErrPathAlreadyHasVersion
 	}
 
-	p := SecretPath(fmt.Sprintf("%s:%d", sp.String(), version))
+	p := SecretPath(sp.String() +":"+ strconv.Itoa(version))
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	p := SecretPath(sp.String() + ":" + strconv.Itoa(version))
```

internals/api/repo.go
```diff
@@ -40,7 +39,7 @@ type Repo struct {
 
 // Path returns the full repository path.
 func (r Repo) Path() RepoPath {
-	return RepoPath(fmt.Sprintf("%s/%s", r.Owner, r.Name))
+	return RepoPath(r.Owner + "/" + r.Name)
```
mackenbach: Consider using the [secretpath](https://godoc.org/github.com/secrethub/secrethub-go/pkg/secretpath) package here. It has convenience functions like `Join` and `AddVersion`. 

Just for your information, the code in this file is a bit old and we don't like it so much. Ideally, we'd move away to something else. I kinda forgot why it's so bad, but I'm sure @SimonBarendse and @jpcoenen have opinions on this.

SimonBarendse: The main problem with the `api` package at the moment is that it is used for the contract between way too many packages.

It's used as an interface between client and server, but it is also used as an interface between the client ([`github.com/secrethub/secrethub-go/pkg/secrethub`](https://pkg.go.dev/github.com/secrethub/secrethub-go/pkg/secrethub?tab=doc)) and consuming applications.

For the path types, we've moved away from using them in the parameters of the client and use primitive types, such as strings instead, because that's easier to understand an d work with (https://github.com/secrethub/secrethub-go/pull/28).

Marton6:
I agree that this refactor does not align with our goals of having high readability and clarity. 
Closed since most changes in this PR negatively impact the readability of the code and (almost) none of them improve it.

PR #177, opened 2020-03-17
Add AWS credential not found error

Marton6:


pkg/secrethub/credentials/sessions/session_aws.go
```diff
@@ -35,13 +38,13 @@ func (s *awsSessionCreator) Create(httpClient *http.Client) (Session, error) {
...
 	if err != nil {
-		return nil, err
+		return nil, handleAWSErr(err)
 	}
 
 	req := api.NewAuthRequestAWSSTS(api.SessionTypeHMAC, region, getCallerIdentityReq)
 	resp, err := httpClient.CreateSession(req)
 	if err != nil {
-		return nil, err
+		return nil, handleAWSErr(err)
```
jpcoenen: I'm not really sure whether we do this correctly in other places, but I think it's best if we use this function directly after calling AWS functionality. That way we avoid placing them at many different places, just to be sure we catch all errors. In this case that would mean placing this function at line 67 and 76 because those are the only calls to the `aws` package.

pkg/secrethub/credentials/sessions/session_aws.go
```diff
@@ -80,3 +83,14 @@ func getCallerIdentityRequest(region string, awsCfg ...*aws.Config) ([]byte, err
...
+func handleAWSErr(err error) error {
+	errAWS, ok := err.(awserr.Error)
+	if ok {
+		if errAWS.Code() == "NoCredentialProviders" {
+			return shaws.ErrNoAWSCredentials
+		}
+		err = errio.Namespace("aws").Code(errAWS.Code()).Error(errAWS.Message())
+	}
+	return err
+}
```
jpcoenen: Should we make the existing function in `internals/aws/errors.go` exported so we can use it here as well and don't have to dupliate it?

pkg/secrethub/credentials/sessions/session_aws.go
Marton6: Good point :+1: I also think that that's the best approach.

pkg/secrethub/credentials/sessions/session_aws.go
Marton6: I wasn't sure about exporting the function but it does seem like a better idea than duplicating code.

internals/aws/service_creator.go
```diff
@@ -34,14 +34,14 @@ type CredentialCreator struct {
...
-		return nil, nil, handleError(err)
+		return nil, nil, HandleError(err)
 	}
 
 	stsSvc := sts.New(sess)
 
 	identity, err := stsSvc.GetCallerIdentity(&sts.GetCallerIdentityInput{})
-	if err != nil {
-		return nil, nil, handleError(err)
+	if err != nil{
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	if err != nil {
```

PR #182, opened 2020-04-10
Malformed credential error

Marton6:
Return a `credential is malformed` error when credential payload is not a valid RSA private key instead of returning an `asn1: trailing data` or similar unhelpful error.

### DIFF
--- a/pkg/secrethub/credentials/encoding.go
+++ b/pkg/secrethub/credentials/encoding.go
@@ -22,9 +22,10 @@ var (
 	ErrInvalidCredentialHeaderField      = errCredentials.Code("invalid_credential_header_field").ErrorPref("invalid header field: %s")
 	ErrCannotDecodeCredentialHeader      = errCredentials.Code("invalid_credential_header").ErrorPref("cannot decode credential header: %v")
 	ErrUnsupportedCredentialType         = errCredentials.Code("unsupported_credential_type").ErrorPref("unsupported credential type: %s")
-	ErrCannotDecodeCredentialPayload     = errCredentials.Code("invalid_credential_header").ErrorPref("cannot decode credential payload: %v")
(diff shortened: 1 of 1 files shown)


pkg/secrethub/credentials/encoding.go
```diff
@@ -226,7 +227,7 @@ type rsaPrivateKeyDecoder struct{}
 func (d rsaPrivateKeyDecoder) Decode(payload []byte) (*RSACredential, error) {
 	key, err := x509.ParsePKCS1PrivateKey(payload)
 	if err != nil {
-		return nil, err
+		return nil, ErrMalformedCredential
```
mackenbach, line 230: Consider making this an `ErrorPref` and wrapping the original error. This will help us quite a lot when debugging a support ticket.

Marton6: Good point.

pkg/secrethub/credentials/encoding.go
```diff
@@ -22,9 +22,10 @@ var (
 	ErrInvalidCredentialHeaderField      = errCredentials.Code("invalid_credential_header_field").ErrorPref("invalid header field: %s")
 	ErrCannotDecodeCredentialHeader      = errCredentials.Code("invalid_credential_header").ErrorPref("cannot decode credential header: %v")
 	ErrUnsupportedCredentialType         = errCredentials.Code("unsupported_credential_type").ErrorPref("unsupported credential type: %s")
-	ErrCannotDecodeCredentialPayload     = errCredentials.Code("invalid_credential_header").ErrorPref("cannot decode credential payload: %v")
+	ErrCannotDecodeCredentialPayload     = errCredentials.Code("invalid_credential_payload").ErrorPref("cannot decode credential payload: %v")
```
SimonBarendse, line 25: Did you intend to include this change? It seems unrelated to this PR? Do we want to change this error code and why?

Marton6: I think it was a mistake. I changed it back

PR #184, opened 2020-04-26
Make setting app info more extensible

florisvdg:
Passing app info to the client is now more extensible, so it can handle:

```
secrethub-go/0.27.0 secrethub-cli/0.37.0 secrethub-circleci-orb/1.0.0
```

Where the third layer comes from `SECRETHUB_APP_INFO_NAME`.

It can now also handle more layers of 'wrapped clients', e.g.: 

```
secrethub-go/0.27.0 secrethub-xgo/0.1.0 secrethub-java/0.2.0 jenkins-secrethub-plugin/1.0.0
```

### Validation
But with greater extensibility, comes greater validation: when using `SECRETHUB_APP_INFO_NAME`, it ignores values if they don't match `[a-zA-Z0-9_-]` and when explicitly setting `WithAppInfo` it actually errors, because as an SDK consumer you'll want to immediately know if (and why) your supplied value didn't come through.

**Note**: there's no validation on the app info _version number_ at the moment, because next to the regular x.y.z versions, there also versions like commit hashes showing up sometimes and it's probably more desirable to have those come through, than have them ignored (or rejected). Agreed? Or should we see if we can find at least some layer of validation there?

### DIFF
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -72,7 +72,7 @@ type Client struct {
 	// These are cached
 	repoIndexKeys map[api.RepoPath]*crypto.SymmetricKey
 
-	appInfo   *AppInfo
+	appInfo   []*AppInfo
 	ConfigDir *configdir.Dir
 }
 
@@ -83,7 +83,7 @@ type AppInfo struct {
 	Version string
 }
 
-func (i AppInfo) userAgentSuffix() string {
+func (i AppInfo) userAgentComponent() string {
 	res := i.Name
 	if i.Version != "" {
 		res += "/" + strings.TrimPrefix(i.Version, "v")
@@ -235,8 +235,8 @@ func (c *Client) DefaultCredential() credentials.Reader {
 
 func (c *Client) userAgent() string {
 	userAgent := userAgentPrefix
-	if c.appInfo != nil {
-		userAgent += " " + c.appInfo.userAgentSuffix()
+	for _, info := range c.appInfo {
+		userAgent += " " + info.userAgentComponent()
 	}
 	osName, err := operatingsystem.GetOperatingSystem()
 	if err != nil {
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -54,12 +54,12 @@ type ClientInterface interface {
 var (
 	errClient = errio.Namespace("client")
 
-	whitelistAppInfoName = regexp.MustCompile("^[a-zA-Z0-9_-]*$")
+	whitelistAppInfoName = regexp.MustCompile("^[a-zA-Z0-9_-]{2,64}$")
 )
 
 // Errors
 var (
-	ErrInvalidAppInfoName = errClient.Code("invalid_app_info_name").Error("name must be set and can only contain alphanumeric, underscore (_), and dash (-) characters")
+	ErrInvalidAppInfoName = errClient.Code("invalid_app_info_name").Error("name must be 2-64 characters long, only alphanumeric, underscore (_), and dash (-)")
 )
 
 // Client is a client for the SecretHub HTTP API.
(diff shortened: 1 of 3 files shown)


pkg/secrethub/client.go
```diff
@@ -139,6 +157,19 @@ func NewClient(with ...ClientOption) (*Client, error) {
...
 
+	appName := os.Getenv("SECRETHUB_APP_INFO_NAME")
+	if appName != "" {
+		appVersion := os.Getenv("SECRETHUB_APP_INFO_VERSION")
+		topLevelAppInfo := &AppInfo{
+			Name:    appName,
+			Version: appVersion,
+		}
+		// Ignore app info from environment variable if name is invalid
+		if err = topLevelAppInfo.ValidateName(); err == nil {
```
jpcoenen, line 168: Do we want to ignore it or just filter out any unsupported characters? The latter could also help with version numbers: if there is something illegal in there, the version number still gets through.

pkg/secrethub/client.go
```diff
@@ -72,7 +80,7 @@ type Client struct {
 	// These are cached
 	repoIndexKeys map[api.RepoPath]*crypto.SymmetricKey
 
-	appInfo   *AppInfo
+	appInfo   []*AppInfo
```
jpcoenen, line 83: I think this can just as well be a `[]AppInfo` -> looks cleaner.

pkg/secrethub/client.go
```diff
@@ -52,6 +53,13 @@ type ClientInterface interface {
 
 var (
 	errClient = errio.Namespace("client")
+
+	whitelistAppInfoName = regexp.MustCompile("^[a-zA-Z0-9_-]*$")
```
jpcoenen, line 57: If we start validating this: do we want a maximum length?

pkg/secrethub/client.go
florisvdg: Hmm, but to make that nice and consistent we'd have to change (and break) `WithAppInfo` too, because that takes `*AppInfo`. So what do we want here?

pkg/secrethub/client.go
florisvdg: For the app name, I think I'd prefer the user agent not getting through over an automatically altered name that may not make any sense.

pkg/secrethub/client.go
florisvdg: Set it to 64 chars now.

pkg/secrethub/client.go
```diff
@@ -54,12 +54,12 @@ type ClientInterface interface {
...
 	errClient = errio.Namespace("client")
 
-	whitelistAppInfoName = regexp.MustCompile("^[a-zA-Z0-9_-]*$")
+	whitelistAppInfoName = regexp.MustCompile("^[a-zA-Z0-9_-]{2,64}$")
 )
 
 // Errors
 var (
-	ErrInvalidAppInfoName = errClient.Code("invalid_app_info_name").Error("name must be set and can only contain alphanumeric, underscore (_), and dash (-) characters")
+	ErrInvalidAppInfoName = errClient.Code("invalid_app_info_name").Error("name must be 2-64 characters long, only alphanumeric, underscore (_), and dash (-)")
```
jpcoenen, line 62: As a reference, 64 characters is this long: `Lorem ipsum dolor sit amet, consectetur adipiscing elit posuerer`. Is there any valid reason for such a long name?

If we make it 32 now, we can later always increase it. In the meantime, it forces people to give the name some thought and make it (relatively) conise (32 is still a lot of characters).

pkg/secrethub/client.go
jpcoenen: Ugh. Yuck. In that case, just leave it.

pkg/secrethub/client.go
florisvdg: The reason I made it longer is that sometimes platforms require a specific naming format for plugins (e.g. https://github.com/jenkinsci/aws-secrets-manager-credentials-provider-plugin). But 64 is indeed somewhat on the high end, so I'll make it 50.

PR #185, opened 2020-05-01
Automatically bump secrethub.ClientVersion on release branch

florisvdg:
image

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -1,0 +1,18 @@
+on:
+  push:
+    branches:
+      - release/v*
+
+jobs:
+  bump-version:
+    name: Bump secrethub.ClientVersion
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Bump version
+        uses: florisvdg/action-version-bump@master
+        with:
+          sed: 's/^\(const ClientVersion = "v\).*\("\)$/\1$VERSION\2/g'
+          file: pkg/secrethub/client_version.go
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  bump-version:
+    name: Bump secrethub.ClientVersion
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Bump version
+        uses: florisvdg/action-version-bump@master
```
SimonBarendse, line 14: Shall we use a pinned version?

florisvdg: I didn't create a release for it yet, but when I do I'll pin it here. But seeing as it's not a critical CI/CD component (you can always set it manually) I'll leave it like this for now.

SimonBarendse: Our branch protection policy doesn't allow a branch to be merged when any of the checks fails. Can we set it optional somehow?

SimonBarendse: Is there a reason not to release a version yet?

florisvdg: It's excluded: 
image

SimonBarendse: Yes, that means if no status is reported at all, that's fine.
However, if the action reports a failure status, the PR isn't allowed to be merged.

florisvdg: Okay pinned it.

PR #191, opened 2020-06-07
Create account along with key in one atomic request

Marton6:
This merge request removes the separate `createAccountKeyRequest` and embeds the account key into the `createCredentialRequest`.
This is done to prevent ending up with unusable accounts if the `createAccountKey` request fails.

### DIFF
--- a/pkg/secrethub/service.go
+++ b/pkg/secrethub/service.go
@@ -67,7 +67,6 @@ func (s serviceService) Create(path string, description string, credentialCreato
 	in := &api.CreateServiceRequest{
 		Description: description,
 		Credential:  credentialRequest,
-		AccountKey:  credentialRequest.AccountKey,
 		RepoMember:  serviceRepoMemberRequest,
 	}
 
--- a/pkg/secrethub/service.go
+++ b/pkg/secrethub/service.go
@@ -54,17 +54,12 @@ func (s serviceService) Create(path string, description string, credentialCreato
 		return nil, errio.Error(err)
 	}
 
-	credentialRequest, err := s.client.createCredentialRequest(credentialCreator.Verifier(), credentialCreator.Metadata())
+	credentialRequest, err := s.client.createCredentialRequest(credentialCreator.Encrypter(), accountKey, credentialCreator.Verifier(), credentialCreator.Metadata())
 	if err != nil {
 		return nil, errio.Error(err)
 	}
 
-	accountKeyRequest, err := s.client.createAccountKeyRequest(credentialCreator.Encrypter(), accountKey)
-	if err != nil {
-		return nil, errio.Error(err)
-	}
-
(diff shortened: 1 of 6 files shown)


Marton6:
This PR is waiting for the next server release.

pkg/secrethub/service.go
```diff
@@ -54,25 +54,20 @@ func (s serviceService) Create(path string, description string, credentialCreato
...
+	serviceRepoMemberRequest, err := s.client.createRepoMemberRequest(repoPath, credentialRequest.AccountKey.PublicKey)
 	if err != nil {
 		return nil, errio.Error(err)
 	}
 
 	in := &api.CreateServiceRequest{
 		Description: description,
 		Credential:  credentialRequest,
-		AccountKey:  accountKeyRequest,
+		AccountKey:  credentialRequest.AccountKey,
```
SimonBarendse, line 70: Aren't we now sending the same account key request twice? I believe we can remove the separate `CreateServiceRequest.AccountKey` altogether. Of course, for backwards compatibility, the server should still accept it though.

jpcoenen:
@SimonBarendse Go ahead and merge it if you're also okay with it.

pkg/secrethub/service.go
```diff
@@ -54,25 +54,19 @@ func (s serviceService) Create(path string, description string, credentialCreato
...
-	serviceRepoMemberRequest, err := s.client.createRepoMemberRequest(repoPath, accountKeyRequest.PublicKey)
+	serviceRepoMemberRequest, err := s.client.createRepoMemberRequest(repoPath, credentialRequest.AccountKey.PublicKey)
 	if err != nil {
 		return nil, errio.Error(err)
 	}
 
 	in := &api.CreateServiceRequest{
 		Description: description,
 		Credential:  credentialRequest,
-		AccountKey:  accountKeyRequest,
```
SimonBarendse, line 75: I believe that error is related to this removal.

Have we changed the API server yet so that it doesn't require this to be set if the account key is supplied via the credential request?

SimonBarendse: We haven't done this yet, so this PR should wait until we've released a version of the API server which accepts the account key either to be supplied at the root of the request (as it currently is) or embedded in the create credential request (as is done after this changeset is applied).

Marton6:
The PR I mentioned this thread implements the required update of the API server.

Marton6:
The change on the API server has been merged.

Marton6:
The API server with the required change has been released.

pkg/secrethub/service.go
Marton6: @SimonBarendse The new version of the API server has been released. Could you test this again and see if you still get any errors?

SimonBarendse (approved):
works now

PR #195, opened 2020-06-23
GCP Linking

jpcoenen:


internals/api/idp_link.go
```diff
@@ -0,0 +1,35 @@
...
+	AuthorizationCode string `json:"authorization_code"`
+}
+
+type IdentityProviderLinkType string
+
+const (
+	IdentityProviderLinkGCP IdentityProviderLinkType = "gcp"
+)
+
+type IdentityProviderLink struct {
```
SimonBarendse, line 20: Maybe we can add some godoc explaining what an identity provider link is and when it is required?


It can use a lot of polishing, but I imagine something like this:
```suggestion
// IdentityProviderLink proves that you have access to a project within the identity provider. This link is required in order to create service accounts that use this identity provider. This is done so that other users cannot claim a SecretHub serivce account that uses the name of your GCP service account, which would keep you from using that name.
// For example, to create GCP service accounts, you first need to prove that you're a member of the GCP project in which you want to create the service account.
type IdentityProviderLink struct {
```

SimonBarendse, line 20: Maybe we can add some godoc explaining what an identity provider link is and when it is required?


It can use a lot of polishing, but I imagine something like this:
```suggestion
// IdentityProviderLink proves that you have access to a project within the identity provider.
// This link is required in order to create service accounts that use this identity provider.
// This is done so that other users cannot claim a SecretHub serivce account that uses the name of your GCP service account, which would keep you from using that name.
//
// For example, to create GCP service accounts, you first need to prove that you're a member of the GCP project in which you want to create the service account.
type IdentityProviderLink struct {
```

pkg/oauthorizer/callback_handler.go
```diff
@@ -0,0 +1,75 @@
...
+func (s CallbackHandler) AuthorizeURL() string {
+	return s.authorizer.AuthorizeLink(s.ListenURL(), s.state)
+}
+
+func (s CallbackHandler) WaitForAuthorizationCode() (string, error) {
+	defer s.listener.Close()
+
+	go func() {
+		err := http.Serve(s.listener, http.HandlerFunc(s.handleRequest))
+		if err != nil && err != http.ErrServerClosed {
```
SimonBarendse, line 53: What happens if the server is closed? will `WaitForAuthorizationCode` ever return?

pkg/secrethub/idp_link.go
```diff
@@ -0,0 +1,70 @@
...
+	client *Client
+}
+
+func (i idpLinkService) GCP() IDPLinkGCPService {
+	return newIDPLinkGCPService(i.client)
+}
+
+type IDPLinkGCPService interface {
+	Create(namespace string, projectID string, authorizationCode string, redirectURI string) (*api.IdentityProviderLink, error)
+	List(namespace string) ([]*api.IdentityProviderLink, error)
```
SimonBarendse, line 28: Should we return an iterator instead, so that we can easily introduce pagination later?

internals/api/idp_link.go
```diff
@@ -33,3 +44,24 @@ type OAuthConfig struct {
 func (c OAuthConfig) Authorizer() oauthorizer.Authorizer {
 	return oauthorizer.NewAuthorizer(c.AuthURI, c.ClientID, c.Scopes...)
 }
+
+// ValidateLinkedID calls the validation function corresponding to the link type and returns the corresponding result.
+func ValidateLinkedID(linkType IdentityProviderLinkType, linkedID string) error {
+	switch linkType {
+	case IdentityProviderLinkGCP:
+		return ValidateGCPServiceAccountEmail(linkedID)
```
SimonBarendse, line 52: typo?

```suggestion
		return ValidateGCPProjectID(linkedID)
```

internals/api/idp_link.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"net/http"
+	"regexp"
+	"time"
+
+	"github.com/secrethub/secrethub-go/pkg/oauthorizer"
+)
+
+var (
+	ErrInvalidIDPLinkType          = errAPI.Code("invalid_idP_link_type").StatusError("invalid IDP link type", http.StatusBadRequest)
```
florisvdg, line 12: Typo?
```suggestion
	ErrInvalidIDPLinkType          = errAPI.Code("invalid_idp_link_type").StatusError("invalid IDP link type", http.StatusBadRequest)
```

internals/api/idp_link.go
jpcoenen: Damn, that's difficult. Gave it a shot. But not sure whether it is very descriptive.

pkg/oauthorizer/callback_handler.go
jpcoenen: Yeah.. this code was bad. New iteration is a bit cleaner.

SimonBarendse: Yes, nice! :ok_hand:

pkg/secrethub/idp_link.go
SimonBarendse: I think we should. Since https://github.com/secrethub/secrethub-go/pull/163, all list functions have been deprecated in favor of iterators. I don't think it makes sense to introduce another list function now, just to deprecate it later.

PR #201, opened 2020-07-07
Add a message to GCP errors without a message

SimonBarendse:
Before https://github.com/secrethub/secrethub-cli/pull/312 the not found error without a message was returned to the user on invalid keyring: `Encountered an error:  (gcp.not_found) `. We should aim never to return these vague errors to the user, but just to be on the safe side for when they do occur, having an actual error message to show would be nice.

### DIFF
--- a/internals/gcp/errors.go
+++ b/internals/gcp/errors.go
@@ -28,17 +28,17 @@ func HandleError(err error) error {
 		switch errGCP.Code {
 		case http.StatusNotFound:
 			if message == "" {
-				message = "404 not found"
+				message = "Response from the Google API: 404 Not Found"
 			}
 			return ErrGCPNotFound.Error(message)
 		case http.StatusForbidden:
 			if message == "" {
-				message = "403 access denied"
+				message = "Response from the Google API: 403 Forbidden"
 			}
 			return ErrGCPAccessDenied.Error(message)
 		case http.StatusConflict:
 			if message == "" {
-				message = "409 conflict"
+				message = "Response from the Google API: 409 Conflict"
 			}
 			return ErrGCPAlreadyExists.Error(message)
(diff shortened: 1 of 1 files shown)


internals/gcp/errors.go
```diff
@@ -24,13 +24,23 @@ var (
 func HandleError(err error) error {
 	errGCP, ok := err.(*googleapi.Error)
 	if ok {
+		message := errGCP.Message
 		switch errGCP.Code {
 		case http.StatusNotFound:
-			return ErrGCPNotFound.Error(errGCP.Message)
+			if message == "" {
+				message = "404 not found"
```
florisvdg, line 31:
```suggestion
				message = "Response from the Google API: 404 Not Found"
```

internals/gcp/errors.go
```diff
@@ -24,13 +24,23 @@ var (
...
+			}
+			return ErrGCPNotFound.Error(message)
 		case http.StatusForbidden:
-			return ErrGCPAccessDenied.Error(errGCP.Message)
+			if message == "" {
+				message = "403 access denied"
```
florisvdg, line 36:
```suggestion
				message = "Response from the Google API: 403 Forbidden"
```

internals/gcp/errors.go
```diff
@@ -24,13 +24,23 @@ var (
...
+			}
+			return ErrGCPAccessDenied.Error(message)
 		case http.StatusConflict:
-			return ErrGCPAlreadyExists.Error(errGCP.Message)
+			if message == "" {
+				message = "409 conflict"
```
florisvdg, line 41:
```suggestion
				message = "Response from the Google API: 409 Conflict"
```

PR #205, opened 2020-07-22
Add AccountService.Me() method

Marton6:
This PR adds the `Me` method to the `AccountService` to allow users of the client to get information about the account used by the client, such as whether it is a service account or a user account.

This method is needed by the CLI to allow treating service accounts and user accounts differently when running the `secrethub account inspect` command.

### DIFF
--- a/pkg/secrethub/account.go
+++ b/pkg/secrethub/account.go
@@ -103,8 +103,8 @@ func (c *Client) getAccountKey() (*crypto.RSAPrivateKey, error) {
 	return c.accountKey, nil
 }
 
-// getMyAccount returns the account of the client itself.
-func (c *Client) getMyAccount() (*api.Account, error) {
+// GetMyAccount returns the account of the client itself.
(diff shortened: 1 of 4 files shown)


jpcoenen:
Task linked: [CU-6wtr69 Teest](https://app.clickup.com/t/6wtr69)

pkg/secrethub/account.go
```diff
@@ -30,6 +32,11 @@ type accountService struct {
 	client *Client
 }
 
+// Me retrieves the account of the client.
```
Marton6, line 35:
```suggestion
// Me retrieves the authenticated account of the client.
```

pkg/secrethub/account.go
```diff
@@ -14,6 +14,8 @@ var (
 
 // AccountService handles operations on SecretHub accounts.
 type AccountService interface {
+	// Me retrieves the account of the client.
```
Marton6, line 17:
```suggestion
	// Me retrieves the authenticated account of the client.
```

PR #206, opened 2020-07-27
Adding CredentialServiceMeService and IDPLinkService to fake client

hculea:
This PR adds the 3 services to the fakeClient so that they can be mocked properly in the tests.

### DIFF
--- a/pkg/secrethub/fakeclient/client.go
+++ b/pkg/secrethub/fakeclient/client.go
@@ -19,7 +19,6 @@ type Client struct {
 	SecretService     *SecretService
 	ServiceService    *ServiceService
 	UserService       *UserService
-	secrethub.ClientInterface
 }
 
 // AccessRules implements the secrethub.Client interface.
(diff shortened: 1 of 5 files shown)


Marton6:
It looks good to me :+1:

pkg/secrethub/fakeclient/client.go
```diff
@@ -10,13 +10,15 @@ import "github.com/secrethub/secrethub-go/pkg/secrethub"
...
+	CredentialService *CredentialService
 	DirService        *DirService
+	IDPLinkService    *IDPLinkService
+	MeService         *MeService
 	OrgService        *OrgService
 	RepoService       *RepoService
 	SecretService     *SecretService
 	ServiceService    *ServiceService
 	UserService       *UserService
-	secrethub.ClientInterface
```
SimonBarendse, line 19: The reason we've put this here is to ensure that `fakeclient.Client` always implements `secrethub.ClientInterface` completely (also if additional methods are added to the interface).

It's okay if we leave it out here, but then let's add `var _ secrethub.ClientIterface = (*Client)(nil)` at the top of this file, so that we still ensure that `fakeclient.Client` implements `secrethub.ClientIterface`.

PR #207, opened 2020-07-27
Expand secret service mock

edif2008:
This PR contains the following features:
 - Fix the `exists` functions which before would always return false
 - Add the possibility to mock `Read` and `ReadString` functions 
 - Implement the `SecretReader` struct to enhance the mocking of the read functions


pkg/secrethub/fakeclient/secret.go
```diff
@@ -12,11 +12,12 @@ type SecretService struct {
...
 	DeleteFunc         func(path string) error
 	GetFunc            func(path string) (*api.Secret, error)
+	ReadFunc           func(path string) (*api.SecretVersion, error)
+	ReadStringFunc     func(path string) (string, error)
+	ExistsFunc         func(path string) (bool, error)
 	WriteFunc          func(path string, data []byte) (*api.SecretVersion, error)
 	ListEventsFunc     func(path string, subjectTypes api.AuditSubjectTypeList) ([]*api.Audit, error)
 	AuditEventIterator *AuditEventIterator
-
-	secrethub.SecretService
```
SimonBarendse, line 19: Same thing as in https://github.com/secrethub/secrethub-go/pull/206 goes for removing this as well.

edif2008: Done

PR #216, opened 2020-08-06
Allow acount key creation with setup code

Marton6:
This PR adds the `WithSetupCode` client option, that makes the client use the supplied setup code for authorizing its requests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
 	google.golang.org/api v0.26.0
 	google.golang.org/genproto v0.0.0-20200331122359-1ee6d9798940
+	google.golang.org/grpc v1.28.0
 )
 
 go 1.13
--- a/pkg/secrethub/credentials/setup_code.go
+++ b/pkg/secrethub/credentials/setup_code.go
@@ -1,0 +1,24 @@
+package credentials
+
+import (
+	"fmt"
+	"github.com/secrethub/secrethub-go/internals/auth"
+	"net/http"
+)
+
+type SetupCode struct {
+	code string
+}
+
+func NewSetupCode(code string) *SetupCode {
+	return &SetupCode{
+		code:code,
+	}
+}
+
+// Authenticate authenticates the given request with a setup code, by providing the "SetupCode" tag and the setup code
+// in the "Authorization" header.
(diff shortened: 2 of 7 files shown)


pkg/secrethub/credentials/setup_code.go
```diff
@@ -0,0 +1,25 @@
...
+		code: code,
+	}
+}
+
+// Authenticate authenticates the given request with a setup code, by providing the "SetupCode" tag and the setup code
+// in the "Authorization" header.
+func (s *SetupCode) Authenticate(r *http.Request) error {
+	r.Header.Set("Authorization", fmt.Sprintf("%s-%s %s", auth.AuthHeaderVersionV1, "SetupCode", s.code))
+	return nil
+}
```
Marton6, line 25: https://github.com/secrethub/secrethub-go/blob/9cc951048891ec17e56d4130cba3bf24fe9c1257/internals/auth/signature.go#L91
`auth.signer.Authenticate` seems to be the method used to set the `Authorization` header in requests, but it cannot be used in this case since it uses a different format for the `Authorization` header (`%s-%s %s:%s` instead of `%s-%s %s`).

The simplest way I found to implement this was to set the `Authorization` header manually like this :point_up: 

Is there a way to improve this code?

SimonBarendse: Actually this is exactly how it was intended to be done. `auth.signer.Authenticate` is only intended to be used for authentication by signing the request. Setup codes use another form of authentication, so they implement their own function.

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
 	google.golang.org/api v0.26.0
 	google.golang.org/genproto v0.0.0-20200331122359-1ee6d9798940
+	google.golang.org/grpc v1.28.0
```
SimonBarendse, line 20: Where does this come from?

Marton6: `go mod tidy` puts it there.

SimonBarendse: Hmm, don't know why, as it's only an indirect dependency via cloud.google.com/go and it's already [in their `go.mod`](https://code.googlesource.com/gocloud/+/refs/tags/v0.56.0/go.mod#27).

I'm getting the same result with `go mod tidy` though.

Let's include it in for now, as I don't see any problems with it, as it's the same pinned version as is defined in the dep.

PR #218, opened 2020-08-14
Remove the with setup code option and make setup code credential provider

Marton6:
As requested here: https://github.com/secrethub/secrethub-cli/pull/332#discussion_r469877042

### DIFF
--- a/pkg/secrethub/credentials/setup_code.go
+++ b/pkg/secrethub/credentials/setup_code.go
@@ -4,11 +4,17 @@ import (
 	"fmt"
 	"net/http"
 
+	httpclient "github.com/secrethub/secrethub-go/pkg/secrethub/internals/http"
+
 	"github.com/secrethub/secrethub-go/internals/auth"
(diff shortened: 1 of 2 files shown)


pkg/secrethub/credentials/setup_code.go
```diff
@@ -4,13 +4,19 @@ import (
...
+
 	"github.com/secrethub/secrethub-go/internals/auth"
 )
 
 type SetupCode struct {
 	code string
 }
 
+func (s *SetupCode) Provide(client *httpclient.Client) (auth.Authenticator, Decrypter, error) {
+	return s, nil, nil
```
jpcoenen, line 19: Can returning `nil` here lead to a nil dereference anywhere or do we already handle this gracefully?

Marton6: We handle it gracefully.

florisvdg (approved):

PR #221, opened 2020-09-10
Resolve and ResolveEnv

edif2008:
This PR contains the following new functionality to the client:
 - Resolve fetches the values of a secret from SecretHub, when the `ref` parameter has the format `secrethub://<path>`
 - ResolveEnv takes a map of environment variables and replaces the values of those which store references of secrets in SecretHub (`secrethub://<path>`) with the value of the respective secret.

### DIFF
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -2,6 +2,7 @@
 // look here to read, write and manage secrets.
 package secrethub
 
+import "C"
 import (
 	"os"
 	"regexp"
@@ -289,4 +290,35 @@ func (c *Client) userAgent() string {
 	userAgent += " (" + osName + "; " + runtime.GOARCH + ")"
 
 	return userAgent
+}
+
+// Resolve fetches the value of a secret, when the `ref` parameter has the
+// format `secrethub://<path>`. Otherwise it returns `ref` unchanged, as an array of bytes.
+func (c *Client) Resolve(ref string) ([]byte, error) {
+	bits := strings.Split(ref, "://")
+	if len(bits) == 2 && bits[0] == "secrethub" {
+		secret, err := c.Secrets().Read(bits[1])
+		if err != nil {
+			return []byte{}, err
+		}
+		return secret.Data, nil
+	}
+	return []byte(ref), nil
+}
+
+// ResolveEnv takes a map of environment variables and replaces the values of those
+// which store references of secrets in SecretHub (`secrethub://<path>`) with the value
+// of the respective secret. The other entries in the map remain untouched.
+func (c *Client) ResolveEnv() (map[string]string, error) {
+	envVars := os.Environ()
+	resolvedEnv := make(map[string]string, len(envVars))
+	for _, value := range envVars {
+		keyValue := strings.Split(value, "=")
+		secretValue, err := c.Resolve(keyValue[1])
+		if err != nil {
+			return map[string]string{}, err
+		}
+		resolvedEnv[keyValue[0]] = string(secretValue)
+	}
+	return resolvedEnv, nil
 }
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -2,7 +2,6 @@
 // look here to read, write and manage secrets.
 package secrethub
 
-import "C"
 import (
 	"os"
 	"regexp"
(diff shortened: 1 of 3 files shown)


pkg/secrethub/client.go
```diff
@@ -290,3 +291,34 @@ func (c *Client) userAgent() string {
 
 	return userAgent
 }
+
+// Resolve fetches the value of a secret, when the `ref` parameter has the
+// format `secrethub://<path>`. Otherwise it returns `ref` unchanged, as an array of bytes.
+func (c *Client) Resolve(ref string) ([]byte, error) {
+	bits := strings.Split(ref, "://")
```
mackenbach, line 298: What happens if you input foobar://path/to/secret ? What is the expected behavior here?

mackenbach, line 298: And also, what happens when you input SECRETHUB://path/to/secret. Is it and should it be case (in)sensitive?

mackenbach (commented):
Saw it pass by, made a few quick comments

pkg/secrethub/client.go
```diff
@@ -290,3 +291,34 @@ func (c *Client) userAgent() string {
...
+		return secret.Data, nil
+	}
+	return []byte(ref), nil
+}
+
+// ResolveEnv takes a map of environment variables and replaces the values of those
+// which store references of secrets in SecretHub (`secrethub://<path>`) with the value
+// of the respective secret. The other entries in the map remain untouched.
+func (c *Client) ResolveEnv() (map[string]string, error) {
+	envVars := os.Environ()
```
mackenbach, line 313: Should the user be in control of what's provided as input here or should it always get the environment?

Use case I'm thinking of is when you have other functions that manipulate the environment, you may want to pass that result in or something like that.

Something along the lines of secrethub.ResolveEnv(os.Environ()). Added benefit is less magic, downside is more typing.

But to be clear, I'm not at all certain about this design decision. It's just an idea.

pkg/secrethub/client.go
```diff
@@ -2,6 +2,7 @@
 // look here to read, write and manage secrets.
 package secrethub
 
+import "C"
```
mackenbach, line 5: Unused import?

pkg/secrethub/client.go
```diff
@@ -290,3 +291,34 @@ func (c *Client) userAgent() string {
 
 	return userAgent
 }
+
+// Resolve fetches the value of a secret, when the `ref` parameter has the
+// format `secrethub://<path>`. Otherwise it returns `ref` unchanged, as an array of bytes.
+func (c *Client) Resolve(ref string) ([]byte, error) {
```
mackenbach, line 297: Consider moving these functions to their own file: env.go. You could even make a separate package called env (which would make env.Resolve and env.ResolveAll) or make a separate 'subservice' called client.Env().Resolve(), or placing these functions on the client.Secrets() subservice.

All in all, these functions are more helper functions that use the client than that they are core client functionality. They cannot exist without a client, so it makes sense to add them, but we can structure them a bit more logically.

pkg/secrethub/client.go
edif2008: Fixed

pkg/secrethub/client.go
edif2008: I made it case insensitive for now.

pkg/secrethub/client.go
edif2008: I moved them in the secretService
They can be called in the following way:
`client.Secrets().Resolve(ref)`
`client.Secrets().ResolveEnv(os.Environ())`

pkg/secrethub/client.go
edif2008: I adjusted the function to accept a parameter as you mentioned.

SimonBarendse (commented):
Hi Eddy,

I think it's really nice to extend the ability to resolve the environment from `secrethub run` in the CLI to the SDK as well!

Do you imagine this being used by SecretHub run as well, to prevent code duplication? Preferably, I think we should have only a single package that defines the `secrethub://` syntax. I think we can easily solve this by using your `ResolveEnv` function here: https://github.com/secrethub/secrethub-cli/blob/d5e091a9492a5a89447142e243206c3fd8d93166/internals/secrethub/env_source.go#L310-L318

Last week we were discussing on Slack whether we should start accepting secret reference syntax on places like `secrethub read` as well. This would be nice, as users might intuitively think of `secrethub://path/to/secret` as a "path", while we currently only accept "path/to/secret" as a "path". If we do want this, I think now would be a good time to add this, before we add a separate function (which we also want to keep backwards compatible) to do just that. In this case, we could add this to the secrets read function:
```Go
if strings.HasPrefix(path, secretReferencePrefix) {
    path = strings.TrimPrefix(path, secretReferencePrefix)
}
```

Before we can do this though, we have to take a holistic look at this feature first; If we start accepting `secrethub://` prefix to read secret paths, we probably should accept it in other places as well.

SimonBarendse:
Do you need `Resolve(ref)` to be public? If not, we could keep it private for now, so that we can delete the function later and move the functionality to the `Read` function, without any breaking changes.

Marton6:
> Do you need `Resolve(ref)` to be public? If not, we could keep it private for now, so that we can delete the function later and move the functionality to the `Read` function, without any breaking changes.

It is needed in the XGO Project, so we want it to be exported.

I think the best course of action would be to implement this in the XGO project and move it to the Go client once we decide on how we want to make it public.

edif2008:
I do think that `secrethub run` with the environments may be something very useful, both for developer and for docker containers.
As for reducing duplicates. if we implement the functionality in the SDK, then in CLI we only need to make the call and print the putput, which simplifies it.
Accepting references as paths can be easily adjusted if we implement it in the SDK. if we want it with a flag, we call the `Resolve` function only then, otherwise it can be always called, since the path will remain untouched if it is not a reference.
In my opinion, the necessary adjustment to bring this new feature are small. Now it is up to you to what extent you want to accept this and how should the user make use of this. In either case, the adjustments are small.

SimonBarendse:
@Marton6 I like that cut :+1: It's relatively little code, so it shouldn't be too much work to implement it in the XGO project I assume?

It will enable us to quickly proceed with the `.NET` SDKs and pick it up for the GO SKD / CLI commands later.

While I do agree, @edif2008 , that it's probably going to be just a couple of slight modifications to the code, I think to make sure it's consistent across all places will require some dedicated attention and time. For example, I expect a discussion about to what extend we allow the `secrethub://` syntax to be used on other types of paths: secrets, directories, probably yes. But do we also allow it for "repo paths" or even workspaces/namespaces/org-names/usernames ? That's a discussion we can postpone for now and focus on the .NET SDK first.

Marton6:
Closed because this feature is not a priority and requires a significant amount of planning to be implemented. This feature will be discussed and implemented in the future.

PR #222, opened 2020-10-14
Improve authentication errors

Marton6:
Currently if any errors occur when providing credentials to the client, these errors are discarded and a `Request not authenticated` error is returned when the first request that requires authentication is made.

This error is vague and can be unhelpful: For example when using the .NET / Go client with a password protected credential, `request not authenticated` does not really address the underlying issue of not being able to use password protected credentials with these clients.

To address this issue, this PR makes the client constructor only ignore `Credential not found` errors, as the only case in which an unauthenticated client is desired is before the user signs up with the CLI. Any other errors (malformed credential, password-protected credential is used with the Go / .NET client, AWS/GCP credential errors) will all be returned when constructing the client.

A new error type has been added to provide a better error message for the error related to password-protected credentials.

**Note that this PR breaks backwards compatibility**, as previously in case of any credential provisioning errors (e.g. malformed credential, aws errors, etc.) an unauthenticated client was returned. After this change the error will be returned instead.

Moreover the credential decoding errors have been wrapped in an error containing the source of the credential, to ease identifying and fixing issues. The error has the following format:
`error decoding credential loaded from '~/.secrethub/credential': ...` or
`error decoding credential loaded from '$SECRETHUB_CREDENTIAL': ...`

### DIFF
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -144,7 +144,7 @@ func NewClient(with ...ClientOption) (*Client, error) {
 		var provider credentials.Provider
 		switch strings.ToLower(identityProvider) {
 		case "", "key":
-			provider = credentials.UseKey(client.DefaultCredential())
+			provider = credentials.UseKey(client.DefaultCredential(), credentials.DefaultKeyDecoder())
 		case "aws":
 			provider = credentials.UseAWS()
 		case "gcp":
@@ -153,7 +153,7 @@ func NewClient(with ...ClientOption) (*Client, error) {
 			return nil, ErrUnknownIdentityProvider(identityProvider)
 		}
 
-		err := client.with(WithCredentials(provider))
+		err = client.with(WithCredentials(provider))
 		// nolint: staticcheck
 		if err != configdir.ErrCredentialNotFound {
 			return nil, err
@@ -265,7 +265,7 @@ func (c *Client) with(options ...ClientOption) error {
 // DefaultCredential returns a reader pointing to the configured credential,
 // sourcing it either from the SECRETHUB_CREDENTIAL environment variable or
 // from the configuration directory.
-func (c *Client) DefaultCredential() credentials.Reader {
+func (c *Client) DefaultCredential() credentials.KeyReader {
 	envCredential := os.Getenv("SECRETHUB_CREDENTIAL")
 	if envCredential != "" {
 		return credentials.FromString(envCredential)
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -266,9 +266,10 @@ func (c *Client) with(options ...ClientOption) error {
 // sourcing it either from the SECRETHUB_CREDENTIAL environment variable or
 // from the configuration directory.
 func (c *Client) DefaultCredential() credentials.KeyReader {
-	envCredential := os.Getenv("SECRETHUB_CREDENTIAL")
+	const credentialEnvVar = "SECRETHUB_CREDENTIAL"
+	envCredential := os.Getenv(credentialEnvVar)
 	if envCredential != "" {
-		return credentials.FromString(envCredential)
+		return credentials.FromEnv(credentialEnvVar)
 	}
 
 	return c.ConfigDir.Credential()
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -155,7 +155,7 @@ func NewClient(with ...ClientOption) (*Client, error) {
 
 		err = client.with(WithCredentials(provider))
 		// nolint: staticcheck
-		if err != configdir.ErrCredentialNotFound {
+		if err != configdir.ErrCredentialNotFound && err != nil {
 			return nil, err
 		} else if err != nil {
 			// TODO: log that default credential was not found.
--- a/pkg/secrethub/client.go
+++ b/pkg/secrethub/client.go
@@ -155,8 +155,10 @@ func NewClient(with ...ClientOption) (*Client, error) {
 
 		err := client.with(WithCredentials(provider))
 		// nolint: staticcheck
-		if err != nil {
-			// TODO: log that default credential was not loaded.
+		if err != configdir.ErrCredentialNotFound {
+			return nil, err
+		} else if err != nil {
+			// TODO: log that default credential was not found.
 			// Do go on because we want to allow an unauthenticated client.
 		}
 	}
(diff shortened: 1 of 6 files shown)


pkg/secrethub/client.go
```diff
@@ -263,10 +265,11 @@ func (c *Client) with(options ...ClientOption) error {
...
 // sourcing it either from the SECRETHUB_CREDENTIAL environment variable or
 // from the configuration directory.
-func (c *Client) DefaultCredential() credentials.Reader {
-	envCredential := os.Getenv("SECRETHUB_CREDENTIAL")
+func (c *Client) DefaultCredential() credentials.KeyReader {
+	const credentialEnvVar = "SECRETHUB_CREDENTIAL"
+	envCredential := os.Getenv(credentialEnvVar)
 	if envCredential != "" {
-		return credentials.FromString(envCredential)
+		return credentials.FromEnv(credentialEnvVar)
```
Marton6, line 272: This change delays reading the credential from the environment variable until later when the credential is actually loaded. This technically breaks backwards compatibility and I can modify it to be (almost) the way it was and still return a nice error.

pkg/secrethub/credentials/readers.go
```diff
@@ -1,52 +1,98 @@
...
+		key, err := decoder.Decode([]byte(os.Getenv(envVarKey)))
+		if err != nil {
+			return Key{}, ErrDecodingCredential{
+				Location: "$" + envVarKey,
+				Err:      err,
+			}
+		}
+		return key, nil
 	})
 }
```
Marton6, line 74: In both of these functions we need to decide whether or not to delay the reading of the credential (file/envvar).

pkg/secrethub/credentials/key.go
```diff
@@ -56,26 +56,37 @@ func (k Key) Export() ([]byte, error) {
...
 }
 
-// ImportKey returns a Key by loading it from the provided credentialReader.
-// If the key is encrypted with a passphrase, passphraseReader should be provided. This is used to read a passphrase
-// from that is used for decryption. If the passphrase is incorrect, a new passphrase will be read up to 3 times.
-func ImportKey(credentialReader, passphraseReader Reader) (Key, error) {
-	bytes, err := credentialReader.Read()
-	if err != nil {
-		return Key{}, err
+type KeyDecoder interface {
```
Marton6, line 59: Please suggest any other more suitable names for this interface. It is currently quite easy to confuse it with the `Decoder` interface which decodes RSA keys. This one decodes the entire credential so maybe something like `CredentialDecoder` would work better, but I'm not sure because it returns the `Key` type.

pkg/secrethub/credentials/encoding.go
```diff
@@ -25,6 +25,7 @@ var (
 	ErrCannotDecodeCredentialPayload     = errCredentials.Code("invalid_credential_header").ErrorPref("cannot decode credential payload: %v")
 	ErrCannotDecodeEncryptedCredential   = errCredentials.Code("cannot_decode_encrypted_credential").Error("cannot decode an encrypted credential without a key")
 	ErrCannotDecryptCredential           = errCredentials.Code("cannot_decrypt_credential").Error("passphrase is incorrect")
+	ErrNeedPassphrase                    = errCredentials.Code("credential_passphrase_required").Error("default credential is password-protected. Configure the client to use service credential by provisioning one through the SECRETHUB_CREDENTIAL environment variable")
```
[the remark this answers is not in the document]
SimonBarendse: The CLI supports `SECRETHUB_CREDENTIAL_PASSPHRASE ` environment variable. Does the SDK already accept that?

If not, I think we should look into starting to accept that, along with accepting `SECRETHUB_CREDENTIAL_PASSPHRASE_CACHE_TTL`.

I think we should do so before introducing this change that requires the passphrase to be supplied.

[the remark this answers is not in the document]
SimonBarendse: The CLI supports `SECRETHUB_CREDENTIAL_PASSPHRASE ` environment variable. Does the SDK already accept that?

If not, I think we should look into starting to accept that, along with accepting `SECRETHUB_CREDENTIAL_PASSPHRASE_CACHE_TTL`.

I think we should do so before introducing this change that requires the passphrase to be supplied.

pkg/secrethub/credentials/key.go
SimonBarendse: See my other comment on not exposing the decoder to the user, as always the same decoder is used.

If we do that, we can remove this interface entirely, or make it private.

pkg/secrethub/client.go
SimonBarendse: Can you provide an example of how this would break backwards compatibility?

SimonBarendse (commented):
Awesome! This will definitely improve the experience for finding the cause of an error. :+1:

About breaking the backwards compatibility, could you provide which functionality breaks with changes introduced? What functionality could be used with an unauthenticated client? Is that only signing up?

pkg/secrethub/credentials/readers.go
```diff
@@ -1,52 +1,98 @@
 package credentials
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 )
 
-// Reader helps with reading bytes from a configured source.
-type Reader interface {
```
SimonBarendse, line 9: Should we keep `Reader` around for a bit for backwards compatibility?

Marton6: Yes, good idea.

pkg/secrethub/client.go
Marton6: Sure:
```go
os.Setenv("SECRETHUB_CREDENTIAL", "ey....") // set the env var to a valid credential
client := secrethub.NewClient()
os.Setenv("SECRETHUB_CREDENTIAL", "invalid-credential") // set the env var to an invalid credential
client.Secrets().ReadString("test/test/test") // read a valid secret
```

This used to work, as when you called `secrethub.NewClient` it took the value of the env var and stored it in the string reader so it didn't matter if you changed it later on. Now it always reads from the environment, so it matters if you change it after calling the constructor.

pkg/secrethub/client.go
```diff
@@ -144,7 +144,7 @@ func NewClient(with ...ClientOption) (*Client, error) {
 		var provider credentials.Provider
 		switch strings.ToLower(identityProvider) {
 		case "", "key":
-			provider = credentials.UseKey(client.DefaultCredential())
+			provider = credentials.UseKey(client.DefaultCredential(), credentials.DefaultKeyDecoder())
```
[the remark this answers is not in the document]
Marton6: Good point! This will significantly simplify the code.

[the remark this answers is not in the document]
Marton6: An alternative decoder is the one which prompts for a passphrase. It is used in the CLI if the credential is password protected.

But you're right, this could be simplified. Maybe we can make the default `credential.UseKey` just use the default key decoder and have a separate function if you want to provide a password. This way we can keep the passwordless keys simple to use and still have an alternative for password protected credentials.

What do you think?

pkg/secrethub/credentials/encoding.go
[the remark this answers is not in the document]
Marton6: I agree. I will look into it.

Marton6:
Closed because changes will be implemented with a new approach in this PR: https://github.com/secrethub/secrethub-go/pull/223

PR #223, opened 2020-11-11
Improve Credential Errors

Marton6:
This PR:
- improves the error returned when the credential is password-protected, but no passphrase reader is provided.
- wraps all errors that occur when loading a credential with a message that includes the credential's source
- adds the `SECRETHUB_CREDENTIAL_PASSPHRASE` environment variable, which is checked if no passphrase reader is provided, but the credential is password-protected

### DIFF
--- a/pkg/secrethub/credentials/encoding.go
+++ b/pkg/secrethub/credentials/encoding.go
@@ -25,7 +25,7 @@ var (
 	ErrCannotDecodeCredentialPayload     = errCredentials.Code("invalid_credential_header").ErrorPref("cannot decode credential payload: %v")
 	ErrCannotDecodeEncryptedCredential   = errCredentials.Code("cannot_decode_encrypted_credential").Error("cannot decode an encrypted credential without a key")
 	ErrCannotDecryptCredential           = errCredentials.Code("cannot_decrypt_credential").Error("passphrase is incorrect")
-	ErrNeedPassphrase                    = errCredentials.Code("credential_passphrase_required").Error("default credential is password-protected. Configure the client to use service credential by provisioning one through the SECRETHUB_CREDENTIAL environment variable")
+	ErrNeedPassphrase                    = errCredentials.Code("credential_passphrase_required").Error("credential is password-protected. Configure the client to use service credential by provisioning one through the SECRETHUB_CREDENTIAL environment variable")
 	ErrMalformedCredential               = errCredentials.Code("malformed_credential").ErrorPref("credential is malformed: %v")
 	ErrInvalidKey                        = errCredentials.Code("invalid_key").Error("the given key is not valid for the encryption algorithm")
 )
--- a/pkg/secrethub/credentials/encoding.go
+++ b/pkg/secrethub/credentials/encoding.go
@@ -25,7 +25,7 @@ var (
 	ErrCannotDecodeCredentialPayload     = errCredentials.Code("invalid_credential_header").ErrorPref("cannot decode credential payload: %v")
 	ErrCannotDecodeEncryptedCredential   = errCredentials.Code("cannot_decode_encrypted_credential").Error("cannot decode an encrypted credential without a key")
 	ErrCannotDecryptCredential           = errCredentials.Code("cannot_decrypt_credential").Error("passphrase is incorrect")
-	ErrNeedPassphrase                    = errCredentials.Code("credential_passphrase_required").Error("credential is password-protected. Configure the client to use service credential by provisioning one through the SECRETHUB_CREDENTIAL environment variable")
+	ErrNeedPassphrase                    = errCredentials.Code("credential_passphrase_required").Error("credential is password-protected. Configure a credential passphrase through the SECRETHUB_CREDENTIAL_PASSPHRASE environment variable or use a credential that is not password-protected")
 	ErrMalformedCredential               = errCredentials.Code("malformed_credential").ErrorPref("credential is malformed: %v")
 	ErrInvalidKey                        = errCredentials.Code("invalid_key").Error("the given key is not valid for the encryption algorithm")
 )
--- a/pkg/secrethub/credentials/encoding.go
+++ b/pkg/secrethub/credentials/encoding.go
@@ -25,6 +25,7 @@ var (
 	ErrCannotDecodeCredentialPayload     = errCredentials.Code("invalid_credential_header").ErrorPref("cannot decode credential payload: %v")
 	ErrCannotDecodeEncryptedCredential   = errCredentials.Code("cannot_decode_encrypted_credential").Error("cannot decode an encrypted credential without a key")
 	ErrCannotDecryptCredential           = errCredentials.Code("cannot_decrypt_credential").Error("passphrase is incorrect")
+	ErrNeedPassphrase                    = errCredentials.Code("credential_passphrase_required").Error("default credential is password-protected. Configure the client to use service credential by provisioning one through the SECRETHUB_CREDENTIAL environment variable")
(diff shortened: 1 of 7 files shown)


pkg/secrethub/credentials/readers.go
```diff
@@ -14,16 +14,17 @@ type Reader interface {
...
 
 // FromEnv returns a reader that reads the contents of an
 // environment variable, e.g. a credential or a passphrase.
 func FromEnv(key string) Reader {
-	return readerFunc(func() ([]byte, error) {
-		return []byte(os.Getenv(key)), nil
+	credential := os.Getenv(key)
+	return newReader("$"+key, func() ([]byte, error) {
+		return []byte(credential), nil
 	})
```
Marton6, line 28: This technically still breaks backwards compatibility, but only if someone relied on the delayed loading of the credential with this function. I think it's not a big issue, but we can create a separate function to maintain this functionality if needed.

pkg/secrethub/credentials/providers.go
```diff
@@ -39,7 +39,12 @@ func (k KeyProvider) Passphrase(passphraseReader Reader) Provider {
 // Provide implements the Provider interface for a KeyProvider.
 func (k KeyProvider) Provide(httpClient *http.Client) (auth.Authenticator, Decrypter, error) {
 	key, err := ImportKey(k.credentialReader, k.passphraseReader)
-	if err != nil {
+	if source, ok := k.credentialReader.(CredentialSource); ok && err != nil {
+		return nil, nil, ErrLoadingCredential{
+			Location: source.Source(),
+			Err:      err,
+		}
+	} else if err != nil {
```
Marton6, line 47: Interestingly, errors regarding the header of the credentials are not wrapped properly, but ones regarding the payload are.

SimonBarendse (approved):
Awesome! Big improvement :smile:

pkg/secrethub/credentials/key.go
```diff
@@ -2,12 +2,22 @@ package credentials
...
 	"github.com/secrethub/secrethub-go/pkg/secrethub/internals/http"
 )
 
+type ErrLoadingCredential struct {
+	Location string
+	Err      error
+}
+
+func (e ErrLoadingCredential) Error() string {
+	return fmt.Sprintf("error loading credential loaded from '%s': %v", e.Location, e.Err)
```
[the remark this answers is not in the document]
Marton6: Sounds good :+1:

pkg/secrethub/credentials/providers.go
```diff
@@ -39,7 +39,12 @@ func (k KeyProvider) Passphrase(passphraseReader Reader) Provider {
...
 		return nil, nil, err
 	}
```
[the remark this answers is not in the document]
Marton6: Yes, it's pretty easy to miss the `err != nil` at the end.

pkg/secrethub/credentials/encoding.go
```diff
@@ -25,6 +25,7 @@ var (
 	ErrCannotDecodeCredentialPayload     = errCredentials.Code("invalid_credential_header").ErrorPref("cannot decode credential payload: %v")
 	ErrCannotDecodeEncryptedCredential   = errCredentials.Code("cannot_decode_encrypted_credential").Error("cannot decode an encrypted credential without a key")
 	ErrCannotDecryptCredential           = errCredentials.Code("cannot_decrypt_credential").Error("passphrase is incorrect")
+	ErrNeedPassphrase                    = errCredentials.Code("credential_passphrase_required").Error("default credential is password-protected. Configure the client to use service credential by provisioning one through the SECRETHUB_CREDENTIAL environment variable")
```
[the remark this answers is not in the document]
Marton6: It seems that this can also be returned from a credential configured through the environment variable.

pkg/secrethub/credentials/providers.go
Marton6: Yes! And the `ImportKey` function is wrapped everywhere except the `account init` command and the `config update passphrase command`. I think it might be a good idea to also wrap the error there

Marton6:
Apparently the errors are still not wrapped correctly. For example in case of an invalid credential supplied through the `SECRETHUB_CREDENTIAL` environment variable.

Marton6:
It seems that the issue with errors not being wrapped correctly was related to the CLI code. I implemented the necessary changes in: https://github.com/secrethub/secrethub-cli/pull/358

pkg/secrethub/credentials/key.go
```diff
@@ -69,8 +79,16 @@ func ImportKey(credentialReader, passphraseReader Reader) (Key, error) {
 		return Key{}, err
 	}
 	if encoded.IsEncrypted() {
+		envPassphrase := os.Getenv("SECRETHUB_CREDENTIAL_PASSPHRASE")
+		if envPassphrase != "" {
+			credential, err := decryptKey([]byte(envPassphrase), encoded)
+			if err != nil {
+				return Key{}, err
```
SimonBarendse, line 86: What error message does this display? Would it help the user to figure out what they should do to solve this error if we would add context indicating we're using a passphrase loaded from the `SECRETHUB_CREDENTIAL_PASSPHRASE` envvar?

pkg/secrethub/credentials/readers.go
SimonBarendse: I don't see any issues in changing the evaluation order.

However, I also don't see any reason to change it. Couldn't we just keep evaluation as is by coding it like this:

```Golang
func FromEnv(key string) Reader {
	return newReader("$"+key, func() ([]byte, error) {
		return []byte(os.Getenv(key)), nil
	})
}
```
```

**Review by SimonBarendse — APPROVED** [review]

File: pkg/secrethub/credentials/key.go (hunk 6 above)
Comment (Marton6) [reply pkg/secrethub/credentials/key.go:86]: Well adding more context to the error would definitely help. I wrapped the error and mentioned the environment variable.

File: pkg/secrethub/credentials/readers.go (hunk 1 above)
Comment (Marton6) [reply pkg/secrethub/credentials/readers.go:28]: Yeah you're right. I implemented your suggestion.

File: pkg/secrethub/credentials/encoding.go [hunk 7]
```diff
@@ -25,6 +25,7 @@ var (
 	ErrCannotDecodeCredentialPayload     = errCredentials.Code("invalid_credential_header").ErrorPref("cannot decode credential payload: %v")
 	ErrCannotDecodeEncryptedCredential   = errCredentials.Code("cannot_decode_encrypted_credential").Error("cannot decode an encrypted credential without a key")
 	ErrCannotDecryptCredential           = errCredentials.Code("cannot_decrypt_credential").Error("passphrase is incorrect")
+	ErrNeedPassphrase                    = errCredentials.Code("credential_passphrase_required").Error("credential is password-protected. Configure the client to use service credential by provisioning one through the SECRETHUB_CREDENTIAL environment variable")
```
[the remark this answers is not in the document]
Marton6: No, it's not the best error message.

What about this?
```suggestion
	ErrNeedPassphrase                    = errCredentials.Code("credential_passphrase_required").Error("credential is password-protected. Configure a credential passphrase through the SECRETHUB_CREDENTIAL_PASSPHRASE environment variable or use a credential that is not password-protected.")
```

SimonBarendse (approved):

pkg/secrethub/credentials/encoding.go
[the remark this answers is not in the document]
jpcoenen: Yes, that's way more useful!

PR #224, opened 2020-11-24
Deprecate wrapping of errors with errio.Error()

jpcoenen:
The wrapping leads to appending "An unexpected error occurred [...]" to many errors, even ones that are relatively simple. The only way to avoid this, is to return all errors as errrio.PublicError; this does not work well togehter with the error paradigm introduced in Go 1.13.

### DIFF
--- a/internals/errio/errors.go
+++ b/internals/errio/errors.go
@@ -105,17 +105,10 @@ func StatusError(err error) error {
 	return UnexpectedStatusError(err)
 }
 
-// Error can be called to on any error to convert it to a PublicError if it is not already.
-// If it is not yet a PublicError, an UnexpectedError is returned
+// Error returns the inputted error.
+// Deprecated: this leads to unwanted returns of UnexpectedErrrors and should therefor not be used anymore.
 func Error(err error) error {
-	if err == nil {
-		return nil
-	}
-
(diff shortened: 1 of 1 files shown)


SimonBarendse (approved):

internals/errio/errors.go
```diff
@@ -105,17 +105,10 @@ func StatusError(err error) error {
...
 func Error(err error) error {
-	if err == nil {
-		return nil
-	}
-
-	if isPublicError(err) || isPublicStatusError(err) {
-		return err
-	}
-	return UnexpectedError(err)
+	return err
```
SimonBarendse, line 111: Not sure if we should change the behavior as well. Should we just deprecate and stop calling the function where we need the changed behavior?

One might depend on the returned type being `UnexpectedError`, as this was explicitly documented and a separate test case exists for this, `TestUnexpectedError`.

jpcoenen:
Had to consider @SimonBarendse thought for a few times. In the end, I think there is an easy solution that keeps the current exported behaviour by leveraging `Unwrap()`. Opened a separate PR for that: https://github.com/secrethub/secrethub-go/compare/feature/unwrap-unexpected-err?expand=1

PR #229, opened 2021-01-28
Handle simultaneous creation of access rules and resources they control

SimonBarendse:
When creating access rules, secrets, secret keys and directories, the SDK first makes a request to fetch what to encrypt. A second request with the encrypted data attached creates the resource. This can lead to conflicts when:
- When creating a new secret, secret key or directory, an access rule giving an additional account access is created after fetching which accounts to encrypt the secret, secret key or directory for.
- When creating a new access rule, a new secret, secret key or directory is created in the directory on which the access rule applies, after fetching what resources to encrypt to create the access rule.

This PR handles the errors when these conflicts arise, fetch the additional data to encrypt and try again. If a conflict still appears after 3 attempts an error is returned. This error contains some additional context compared to the error returned by the API server.


pkg/secrethub/acl.go
```diff
@@ -218,50 +228,113 @@ func (s accessRuleService) create(path api.BlindNamePath, permission api.Permiss
...
-		if err != nil {
-			return nil, errio.Error(err)
+		accessRule, err := s.client.httpClient.CreateAccessRule(blindName, accountName, in)
+		if err != api.ErrNotEncryptedForAccounts {
+			return accessRule, err
 		}
+		if tries >= missingMemberRetries {
+			return nil, fmt.Errorf("cannot create access rule: resources controlled by the access rule are simultaneously being created; you may try again")
+		}
+		tries++
```
jpcoenen, line 256: What if `err == nil`?

jpcoenen (commented):
Mostly looks very nice :+1: Only one question about an error-check.

pkg/secrethub/acl.go
SimonBarendse: This is another example, just like the other PR reviewed yesterday, where the negated comparison for a specific error also includes the nil case for the error:

```go
if err != api.ErrNotEncryptedForAccounts {
	return accessRule, err
}
```

it's exactly equal to:
```go
if err == nil {
	return accessRule, nil
}
if err != api.ErrNotEncryptedForAccounts {
	return nil, err
}
```

Like before, I'll change to the latter implementation, as after the previous discussion we agreed to value the explicitness (and the readability that comes with that) of the latter over the conciseness and the negligible performance gain (of having just a single error check instead of two) of the former.

jpcoenen (approved):

PR #234, opened 2021-02-05
Fix error handling of credential loading

jpcoenen:
There were two cases where either no error or an useless error were returned in the process of loading a credential:

- If no passphrase was provided when using the SDK, this error would get ignored altogether. This is fixed by 8a82982.
- After this was fixed, not setting a passphrase would lead to an "invalid passphrase" error instead of a "no passphrase provided error. This is fixed by 1a490aa (`passphraseReader` is now `nil` again in all cases that no passphrase is provided)
- If an invalid passphrase was provided through the environment, a crypto error would be returned. This is fixed by 9f74af2.

### DIFF
--- a/pkg/secrethub/credentials/key.go
+++ b/pkg/secrethub/credentials/key.go
@@ -85,6 +85,9 @@ func ImportKey(credentialReader, passphraseReader Reader) (Key, error) {
 		if envPassphrase != "" {
 			credential, err := decryptKey([]byte(envPassphrase), encoded)
 			if err != nil {
+				if crypto.IsWrongKey(err) {
+					err = ErrCannotDecryptCredential
+				}
 				return Key{}, fmt.Errorf("decrypting credential with passphrase read from $%s: %v", credentialPassphraseEnvVar, err)
 			}
 			return Key{key: credential}, nil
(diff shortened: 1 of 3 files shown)


pkg/secrethub/credentials/key.go
```diff
@@ -85,6 +85,9 @@ func ImportKey(credentialReader, passphraseReader Reader) (Key, error) {
 		if envPassphrase != "" {
 			credential, err := decryptKey([]byte(envPassphrase), encoded)
 			if err != nil {
+				if crypto.IsWrongKey(err) {
+					err = ErrCannotDecryptCredential
+				}
 				return Key{}, fmt.Errorf("decrypting credential with passphrase read from $%s: %v", credentialPassphraseEnvVar, err)
```
SimonBarendse, line 91: Is it useful to include the source of the passphrase in this error message as well?

SimonBarendse (approved):

pkg/secrethub/credentials/key.go
jpcoenen: I'm not sure if I get you correctly. Because I think that already happens.

SimonBarendse: Sorry, all is good :+1:

I misread this line `err = ErrCannotDecryptCredential` and thought the error was returned there. :see_no_evil:

SimonBarendse (approved):

PR #236, opened 2021-03-31
Improve performance of creating access rules

hculea:
In order to check if the optimisation is effectual, I have run the following script, which populated the priorly created `optimization-test` directory with 1000 secrets:

```go
u, err := secrethub.NewClient()
if err != nil {
	fmt.Print(err.Error())
}

iterations := 1000
for i := 0; i < iterations; i++ {
	_, err := u.Secrets().Write(fmt.Sprintf("<SECRETHUB NAMESPACE>/optimization/optimization-test/secret%d", i), []byte(fmt.Sprintf("value%d", i)))
	if err != nil {
		fmt.Print(err.Error())
	}
}
```

I have then replaced the CLI's SDK dependency by pasting the following in CLI project's `go.mod`:
```
replace (
    github.com/secrethub/secrethub-go => <RELATIVE PATH TO secrethub-go FOLDER>  // in my case, ../secrethub-go
)
```
and then I rebuilt the CLI.

NB, I was on the `feature/refactor-and-test-acl` branch on the SDK.

The effect of the pull request can be observed by comparing the running times of an `acl set` command run on a directory with a large number of secrets (such as the one created by the script above), before and after the change, which, in my experience, differed by quite a large factor. 

If anyone considers this necessary, I will measure the actual times for both the before and after cases, making use of the `duration` library.

### DIFF
--- a/pkg/secrethub/acl.go
+++ b/pkg/secrethub/acl.go
@@ -1,7 +1,9 @@
 package secrethub
 
 import (
+	"context"
 	"fmt"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/secrethub/secrethub-go/internals/crypto"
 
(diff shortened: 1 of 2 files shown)


SimonBarendse (approved):
Sweet stuff! :ok_hand:

I've tested this on a huge repo (2000 secrets). On `develop` it takes a whopping 231s to create an access rule for the root, with these changes applied it takes 28s. Although that's still a long while (the encryption itself also takes long), it's a huge improvement! Thanks for adding this! :blue_heart:

I've pushed a couple of changes to the test suite. Could you have a look at them? I've added some explanation in the commit messages. If you agree with what I've done, this is ready to go :rocket:

hculea:
Thanks for the touch ups! Everything looks good to me too! I'll merge now.

PR #237, opened 2021-03-31
Account delete

Marton6:
This PR adds a `Delete` method to the account service, that deletes the currently authenticated account.

To prevent users from calling this method accidentally, the method requires you to pass your own account id as parameter.


SimonBarendse (commented):
Looks all good :+1:

We should wait with merging this until server-side changes are released.

SimonBarendse (approved):

